PR #108, opened 2019-07-12
Fix: paging for API Operations list

### LINKED ISSUE #54: Pagination in API Operation
In API page max 20 operations can be list out and will not show if we have more than 20 operations, then we are not seeing any pagination for Page Contract.

Is there any way around or it will be published as new feature?

ygrik:
close #54


src/components/operations/runtime/operation-list/operation-list.ts
```diff
@@ -105,6 +109,27 @@ export class OperationList {
...
+        this.lastPage(pageOfOperations);
+        const current = this.operations();
+        if (current && current.length > 0) {
+            this.operations.push(...pageOfOperations.value);
+        } else {
+            this.operations(pageOfOperations.value);
+        }
+    }
+
+    public async loadMore() {
```
azaslonov: : Promise<void>?

PR #190, opened 2019-10-04
Widget: Products list

### LINKED ISSUE #185: New widget: list of Products

ygrik:
Closes #185 

image

### DIFF
--- a/src/components/products/product-list/ko/runtime/product-list.ts
+++ b/src/components/products/product-list/ko/runtime/product-list.ts
@@ -103,19 +103,12 @@ export class ProductList {
 
         try {
             const itemsPage = await this.productService.getProductsPage(query);
-            const contracts = itemsPage.value;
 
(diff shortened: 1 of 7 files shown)


src/components/products/product-list/ko/runtime/product-list.ts
```diff
@@ -28,32 +37,86 @@ export class ProductList {
...
+            const searchItems = [];
 
-            if (productId) {
-                this.selectedProductId(`/products/${productId}`);
-            }
-            else if (products.length > 0) {
-                productId = products[0].id;
-                this.selectProduct(productId);
+            if (contracts) {
+                contracts.map(item => searchItems.push(new Product(item)));
```
azaslonov: I would move this inside service method, so that it returned models rather than contracts:
`contracts.map(item => searchItems.push(new Product(item)));`

PR #225, opened 2019-10-22
Add Terms Of Use to Sign Up

### LINKED ISSUE #228: Add terms of use to the user sign up

RupengLiu:
close #228
unchecked the error
image
checked the error
image
show terms of use
image

### DIFF
--- a/src/components/users/user-signup/ko/runtime/user-signup.ts
+++ b/src/components/users/user-signup/ko/runtime/user-signup.ts
@@ -26,7 +26,6 @@ export class UserSignup {
     public readonly firstName: ko.Observable<string>;
     public readonly lastName: ko.Observable<string>;
     public readonly isUserRequested: ko.Observable<boolean>;
-    public readonly isConsentRequired: ko.Observable<boolean>;
     public readonly showTerms: ko.Observable<boolean>;
     public readonly consented: ko.Observable<boolean>;
     public readonly showHideLabel: ko.Observable<string>;
@@ -51,10 +50,19 @@ export class UserSignup {
         this.working = ko.observable(false);
         this.captcha = ko.observable();
         this.delegationUrl = ko.observable();
+        this.termsEnabled = ko.observable();
     }
 
     @Param()
     public termsOfUse: ko.Observable<string>;
+
+    @Param()
+    public isConsentRequired: ko.Observable<boolean>;
+
+    @Param()
+    public termsEnabled: ko.Observable<boolean>;
+
+    @Param()
     public delegationUrl: ko.Observable<string>;
 
     /**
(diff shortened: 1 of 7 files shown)


src/components/users/user-signup/ko/runtime/user-signup.ts
```diff
@@ -84,10 +84,9 @@ export class UserSignup {
         }
 
         const settings = {
```
azaslonov: A couple of things here:
1. We need to take into account that terms of use may not be specified at all.
2. Settings `userRegistrationTermsEnabled` and `userRegistrationTermsConsentRequired` should also come from configuration.

RupengLiu: k~, will fix that

src/components/users/user-signup/ko/runtime/user-signup.ts
```diff
@@ -84,20 +92,15 @@ export class UserSignup {
...
-            // TODO: Registration terms could be rendered at publish time
-            userRegistrationTerms: "Test userRegistrationTerms!!!",
-            userRegistrationTermsEnabled: false,
-            userRegistrationTermsConsentRequired: false
+            userRegistrationTerms: this.termsOfUse(),
+            userRegistrationTermsEnabled: this.termsEnabled(),
+            userRegistrationTermsConsentRequired: this.isConsentRequired()
         };
-
         this.tenantSettings = settings as TenantSettings;
```
ygrik: I do not see any good reasons to keep this **tenantSettings **
do you really need it?

RupengLiu: sure i will delete it

src/components/users/user-signup/ko/userSignupViewModelBinder.ts
```diff
@@ -21,9 +30,14 @@ export class UserSignupViewModelBinder implements ViewModelBinder<UserSignupMode
         delegationParam[DelegationParameters.ReturnUrl] =  "/";
 
         const delegationUrl = await this.tenantService.getDelegationUrl(DelegationAction.signIn, delegationParam);
-        if (delegationUrl) {
-            viewModel.delegationConfig(JSON.stringify({ delegationUrl: delegationUrl}));
-        }
+        const termsOfService = await this.getTermsOfService();
+        const runtimeConfig = JSON.stringify({
```
ygrik: You should not add config object if there is no settings

src/components/users/user-signup/ko/userSignupViewModelBinder.ts
```diff
@@ -21,9 +30,14 @@ export class UserSignupViewModelBinder implements ViewModelBinder<UserSignupMode
         delegationParam[DelegationParameters.ReturnUrl] =  "/";
 
         const delegationUrl = await this.tenantService.getDelegationUrl(DelegationAction.signIn, delegationParam);
-        if (delegationUrl) {
-            viewModel.delegationConfig(JSON.stringify({ delegationUrl: delegationUrl}));
-        }
+        const termsOfService = await this.getTermsOfService();
```
ygrik: Also please use try-catch for MAPI calls

PR #227, opened 2019-10-22
Full delegation for products and authentication

ygrik:
added delegation for subscribe, unsubscribe, change password, change profile, close account

Closes https://github.com/Azure/api-management-developer-portal/issues/212


src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
```diff
@@ -123,23 +128,55 @@ export class ProductSubscribe {
         this.limitReached(limitReached);
     }
 
+    private async isDelegation(userId: string, productId: string): Promise<boolean> {
```
azaslonov: maybe "isDelegationEnabled"?

PR #237, opened 2019-10-27
Added registration step for non-existent users authenticated with AAD (B2C)

### LINKED ISSUE #204: Azure AD Login - Issue
**Describe the bug**
Ive configured api management "Azure Active Directory" identity provide. Im not able to login  into  preview developer portal. After sign-in process its redirect to signin page doesnt process the token and profile page was empty.  but when i try the same process in old portal its works. 

And also menu still doesnt show the logged in user name . i was expecting after sign-in process menu should show profile name and sign out link ( both are missing in the new preview portal ) .

it would be great if you could address this issues.

When i try to debug the authentication service code. Getting below error in the aadservices.ts - > exchangeIdToken

Authentication failed. Header with access token is missing.

azaslonov:
Closes #204.

### DIFF
--- a/src/services/aadService.ts
+++ b/src/services/aadService.ts
@@ -24,7 +24,9 @@ export class AadService {
             const jwtToken = Utils.parseJwt(idToken);
 
             setTimeout(async () => {
-                const email = prompt("Enter your email address");
+                const email = jwtToken.email
+                    ? jwtToken.email
+                    : prompt("Enter your email address");
 
                 const user: UserPropertiesContract = {
                     firstName: jwtToken.given_name,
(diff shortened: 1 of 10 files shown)


src/services/aadService.ts
```diff
@@ -14,10 +16,49 @@ export class AadService {
         private readonly settingsProvider: ISettingsProvider
     ) { }
 
+    private async createUser(idToken: string, type: string): Promise<HttpHeader> {
+        const managementApiUrl = await this.settingsProvider.getSetting<string>("managementApiUrl");
+        const managementApiVersion = await this.settingsProvider.getSetting<string>("managementApiVersion");
```
ygrik: are they from Constants.SettingNames?

azaslonov: Yep. replaced.

PR #254, opened 2019-10-30
Reports UI fix

RupengLiu:
image
image


src/components/reports/ko/runtime/reports.ts
```diff
@@ -359,6 +367,10 @@ export class Reports {
         });
 
         this.reportByOperation(viewModels);
+        } 
+        catch (error) {
+            console.log(error);
```
azaslonov: Looks like we don't handle the exception here, so let's just remove `try catch` block.

PR #353, opened 2019-12-05
Allow the operation-list group by tag to be enabled by default

j12n:
Feature enhancement to allow the group by tag toggle to be enabled by default on the operation list page.


neterium:
Sorry for asking, but where (in Azure) can we toggle this ? I do not see any option to enable the operation-list group by feature by default.
Thanks,

azaslonov:
You can find this option in the editor of the "Operation list" widget:
image

PR #416, opened 2020-01-14
display values in dropdown for query and headers params in console

### LINKED ISSUE #385: No dropdown list for enum parameters
Hello,

I have observed a small difference in functionnality between the old and the new developer portal.

## Bug description

The new developer portal does not provide a dropdown list for enumeration parameters in header.

## Reproduction steps

1. Import the [swagger.txt](https://github.com/Azure/api-management-developer-portal/files/3998410/swagger.txt) in Azure API Management (it is a yaml definition so you may rename the file extension properly).
2. In the old development portal, we have a dropdown list for the header Gender in the operation POST Add a new pet to the store. 
Developer portal legacy
3. In the new developer portal, when we try the operation, we do not get the dropdown list for the header. 
New developer portal

## Expected behavior

We expect a dropdown list in the new developer portal, that makes it easier for developers trying our APIs.

Managed

## Environment

 - Browser: Google Chrome, Edge

ygrik:
close #385

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/operation-console.ts
+++ b/src/components/operations/operation-details/ko/runtime/operation-console.ts
@@ -149,8 +149,23 @@ export class OperationConsole {
 
         await this.loadSubscriptionKeys();
 
+        this.trackOptionsChanges();
         this.updateRequestSummary();
         this.working(false);
+    }
+
+    private trackOptionsChanges(): void {
+        if (this.consoleOperation()) {
+            const operation = this.consoleOperation();
+            const queryParams = operation.request.queryParameters();
+            if (queryParams) {
+                queryParams.map(p => p.options.length > 0 && p.value.subscribe(() => this.updateRequestSummary()));
+            }
+            const headers = operation.request.headers();
+            if (headers) {
+                headers.map(p => p.options.length > 0 && p.value.subscribe(() => this.updateRequestSummary()));
+            }
+        }
     }
 
     private setSoapHeaders(): void {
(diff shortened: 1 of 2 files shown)


src/components/operations/operation-details/ko/runtime/operation-console.ts
```diff
@@ -149,10 +149,25 @@ export class OperationConsole {
 
         await this.loadSubscriptionKeys();
 
+        this.trackOptionsChanges();
         this.updateRequestSummary();
         this.working(false);
     }
 
+    private trackOptionsChanges(): void {
```
azaslonov: I guess `templateParameters` are missing.

src/components/operations/operation-details/ko/runtime/operation-console.ts
```diff
@@ -149,10 +149,25 @@ export class OperationConsole {
...
+        if (this.consoleOperation()) {
+            const operation = this.consoleOperation();
+            const queryParams = operation.request.queryParameters();
+            if (queryParams) {
+                queryParams.map(p => p.options.length > 0 && p.value.subscribe(() => this.updateRequestSummary()));
```
azaslonov: doesn't this binding above do the same? `data-bind="event: { keyup: $component.updateRequestSummary }`

ygrik: is keyup event triggers on mouse events?

azaslonov: no, in this case maybe it's not needed in markup?

PR #489, opened 2020-02-14
Add automated migrate script

michaeltnguyen:
This PR adds `migrate.js`, which supports fully automated migrations between developer portals.  When used for a managed portal, the storage connection strings are automatically retrieved from the management API.  Users can specify SAS tokens directly, or they may input their `identifier` and `key`, and the script will generate a SAS token on their behalf.

Managed usage with autogenerated tokens:
`node migrate.js --autoPublish --sourceName <my-source> --sourceId <id> --source-key <key> --destName <my-destination> --destId <id> --destKey <key>`

Self-hosted usage:
`node migrate.js --self-hosted --autoPublish --sourceName <my-source> --sourceToken <token> --sourceStorage <storage> --destName <my-destination> --destToken <token> --destStorage <storage>`

Those examples are also viewable by running with the `--help` option.

Some notes:
- No new dependencies are added.  (`moment` and `yargs` are already pulled in via other dependencies).
- I opted to not do an interactive mode at this time, since it would require even more inputs (subscriptionId, resource group)
- I tested this on two developer portals to ensure it worked.  (running on mac and windows)

### DIFF
--- a/scripts/migrate.js
+++ b/scripts/migrate.js
@@ -1,0 +1,227 @@
+/**
+ * This script automates deployments between developer portal instances.
+ * In order to run it, you need to:
+ * 
+ * 1) clone the api-management-developer-portal repository
+ * 2) npm install in the root of the project
+ * 3) install az-cli (https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest)
+ * 4) run this script with a valid combination of arguments
+ * 
+ * If you run with the --selfHosted flag, you are expected to supply a sourceStorage and destStorage parameter.
+ * 
+ * You can specify the SAS tokens directly (via sourceToken and destToken), or you can supply an identifier and key,
+ * and the script will generate tokens that expire in 1 hour. (via sourceId, sourceKey, destId, destKey)
+ */
+
+const https = require('https');
+const mkdirSync = require('fs').mkdirSync;
+const execSync = require('child_process').execSync;
+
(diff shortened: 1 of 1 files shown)


azaslonov:
Hi Michael, that's an awesome script! Thanks a lot! Could you please sign CLA? That's the requirement that we can't overcome.

michaeltnguyen:
Just signed.

It also just occurred to me that `autoPublish` might not work with self-hosted portals as-written.  There's a "management endpoint" (management.azure-api.net) and a portal endpoint (developer.azure-api.net).  Do self-hosted portals use the same management endpoint, but specify their own portal endpoint?

mikebudzynski:
> It also just occurred to me that autoPublish might not work with self-hosted portals as-written. > There's a "management endpoint" (management.azure-api.net) and a portal endpoint (developer.azure-api.net). Do self-hosted portals use the same management endpoint, but specify their own portal endpoint?

@michaeltnguyen Apologies for a late response. You're right - the autopublish will not work for the self-hosted portal.

You can take a look at how we did this here: https://github.com/Azure/api-management-developer-portal/blob/master/scripts/deploy-default.bat

For the self-hosted portal, publishing requires running the publishing step locally and then uploading the files to a blob storage account.

Nevertheless, the script is extremely useful and thanks for your contribution! Most importantly, it's a simple user experience for the managed portal users. I'll play with it in the coming days and see if we can easily do the publish for the self-hosted portal with it.

PR #530, opened 2020-03-06
Revise the migration script for custom domains, add comments, remove old script

mikebudzynski:


mikebudzynski:
@michaeltnguyen I have adapted the script to support custom domains in the managed version. For now, the publishing functionality won't be available for the self-hosted version, because it requires:

- Changing the configuration files `config.design.json`, `config.publish.json`, `config.runtime.json`
- Publishing the portal locally
- Uploading the generated files to a desired hosting (e.g., storage account)

michaeltnguyen:
makes sense.  Thanks for taking care of this!

mikebudzynski:
I've also revised the migration instructions in wiki: https://github.com/Azure/api-management-developer-portal/wiki/Migrate-portal-between-services

PR #621, opened 2020-05-19
Improved handling of AAD and AAD B2C signin/signup process.

azaslonov:


src/services/aadService.ts
```diff
@@ -77,16 +92,16 @@ export class AadService {
             scopes: ["openid", "email", "profile"]
         };
 
-        try {
+        // try {
```
ygrik, line 95: do we need it?

azaslonov: Nope :)

PR #680, opened 2020-06-04
Fixed reset password

### LINKED ISSUE #588: Can't reset password on self-hosted portal
**NOT THE SAME AS #587** 

## Bug description

When self-hosting the portal a user is unable to reset their password. The link that gets sent in the email works perfectly well for the managed version of the portal, but not for the self-hosted version. 

The developer tools in Chrome show that a request to the management api fails with a 401 Unauthorized error.

## Reproduction steps

1. Go to the homepage
2. Click on 'Sign In'
3. Click on 'Forgot your password?'
4. Fill in and Click 'Request reset'
5. Open link in email
6. **An error message is displayed 'Activate user error: You're not authorized.' above the form.**
7. Fill in the form with a new password
8. Click 'Reset' button
9. **A different error message is displayed 'Server error. Unable to send request. Please try again later.'**

## Expected behavior

The process for resetting a password when using a self-hosted version of the portal is identical to the process in the managed version of the portal.

Self-hosted

commit #578, [43441ba9b2abcc952cfbaa479798d23e79dff25f]

## Environment

 - Operating system: [Azure Blob Storage]
 - Browser: [Google Chrome]
 - Version: [latest]

## Additional context

The email template is configured to generate the link as:

**

> https://[portal-url]/confirm-password?$ConfirmQuery

**

I think the bug has been introduced in commit #516 which tried to resolve issue #460. 

In particular **the first error** message is thrown on **line 86 in confirm-password.ts**. The error message is thrown when the application calls **usersService.activateUser(queryParams);**

There are several issues with having this functionality in the **initialize()** method, because the same code tries to handle different cases:

1. Confirm a user
2. Reset a user's password

The query parameters are not the same, meaning if the **usersService.activateUser** was executed successfully the identity will be set to 'null' rendering the user unable to login even if they reset their password. (The identity in this case is the user's login name) 

> public async activateUser(parameters: URLSearchParams): Promise<void> {
>         const userId = parameters.get("userid");
>         const ticket = parameters.get("ticket");
>         const ticketId = parameters.get("ticketid");
>         const identity = parameters.get("identity");
>         const requestUrl = `/users/${userId}/identities/Basic/${identity}`;
>         const token = `Ticket id="${ticketId}",ticket="${ticket}"`;
>         
>         await this.mapiClient.put<void>(requestUrl, [{ name: "Authorization", value: token }], {});
>     }

The **$ConfirmQuery** available in the Notification templates in Azure only provides the first three parameters. My guess is that this method should not be called when the user is just trying to reset their password. 

**The second error** message is thrown on line 129 in the **confirm-password.ts** file. My guess is there is problem with the **usersService.updatePassword** method. It passes a null object as a header and it makes sense the management API would refuse to accept the request.

> public async updatePassword(userId: string, newPassword: string): Promise<void> {
>         await this.mapiClient.patch(userId, undefined, { password: newPassword });
>     }
>

ygrik:
close #588

### DIFF
--- a/src/services/usersService.ts
+++ b/src/services/usersService.ts
@@ -56,6 +56,20 @@ export class UsersService {
         }
     }
 
+    public getTokenFromTiketParams(parameters: URLSearchParams): string {
+        const ticket = parameters.get("ticket");
+        const ticketId = parameters.get("ticketid");
+        const token = `Ticket id="${ticketId}",ticket="${ticket}"`;
+
+        return token;
+    }
+
+    public getUserIdFromParams(parameters: URLSearchParams): string {
+        const userId = parameters.get("userid");
+
+        return userId ? `/users/${userId}` : undefined;
+    }
+
     public async activateUser(parameters: URLSearchParams): Promise<void> {
         const userId = parameters.get("userid");
         const ticket = parameters.get("ticket");
@@ -67,8 +81,9 @@ export class UsersService {
         await this.mapiClient.put<void>(requestUrl, [{ name: "Authorization", value: token }], {});
     }
 
-    public async updatePassword(userId: string, newPassword: string): Promise<void> {
-        await this.mapiClient.patch(userId, undefined, { password: newPassword });
+    public async updatePassword(userId: string, newPassword: string, token: string): Promise<void> {
+        const heasers = [{ name: "Authorization", value: token }];
+        await this.mapiClient.patch(userId, heasers, { password: newPassword });
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/services/usersService.ts
```diff
@@ -56,6 +56,20 @@ export class UsersService {
         }
     }
 
+    public getTokenFromTiketParams(parameters: URLSearchParams): string {
```
BobbyCGD, line 59: Typo in method name, 'Ticket' not 'Tiket'

ygrik: Thank you! good catch :) will fix it

PR #697, opened 2020-06-10
fixed typo and added check if token is not provided for update password

ygrik:


src/services/usersService.ts
```diff
@@ -82,7 +82,10 @@ export class UsersService {
     }
 
     public async updatePassword(userId: string, newPassword: string, token: string): Promise<void> {
-        const heasers = [{ name: "Authorization", value: token }];
+        const heasers = []
```
azaslonov, line 85: Typo.

PR #836, opened 2020-08-19
Fixed issue with gateway info initialization. Fixed few issues with object and array representations.

### LINKED ISSUE #531: Array definitions are not displayed in the Developer portal
## Bug description
When the API contains an array of objects, the item description of the array is not displayed in the Developer Portal.

## Reproduction steps
I have this definition in my Swagger:
`{
    "type": "object",
    "properties": {
        "Products": {
            "type": "array",
            "items": {
                "type": "object",
                "description": "products of the company",
                "properties": {
                    "productName": {
                        "type": "string",
                        "description": "name of the product"
                    }
                },
                "required": [
                    "productName"
                ]
            }
        }
    }
}`

I have an operation with a response associated with the type of this definition.

Unfortunately, the portal did not represent the definition of my object because it is an array:
image

## Expected behavior
Expected behavior is the same as the legacy portal as shown below:
image

azaslonov:
- Fixed gateway info initialization;
- Fixed case with empty object property in schema definition;
- Added flattening of arrays with object-like items in schema definition (Closes #531);

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/operation-details.ts
+++ b/src/components/operations/operation-details/ko/runtime/operation-details.ts
@@ -97,7 +97,6 @@ export class OperationDetails {
         this.selectedOperationName(operationName);
 
         if (apiName) {
-            await this.loadGatewayInfo(apiName);
             await this.loadApi(apiName);
         }
 
@@ -137,6 +136,8 @@ export class OperationDetails {
         if (!api) {
             return;
         }
+
+        await this.loadGatewayInfo(apiName);
 
         this.api(api);
 
--- a/src/models/typeDefinition.ts
+++ b/src/models/typeDefinition.ts
@@ -263,16 +263,22 @@ export class TypeDefinitionObjectProperty extends TypeDefinitionProperty {
 
                                 if (propertySchemaObject.items.$ref) {
                                     arrayProperty.type = new TypeDefinitionPropertyTypeArrayOfReference(getTypeNameFromRef(propertySchemaObject.items.$ref));
+                                    props.push(arrayProperty);
+                                }
+                                else if (propertySchemaObject.items.properties) { 
+                                    const objectProperty = new TypeDefinitionObjectProperty(propertyName, propertySchemaObject.items, isRequired, true);
+                                    const flattenObjects = this.flattenNestedObjects(objectProperty, propertyName + "[]");
+                                    props.push(...flattenObjects);
                                 }
                                 else if (propertySchemaObject.items.type) {
                                     arrayProperty.type = new TypeDefinitionPropertyTypeArrayOfPrimitive(propertySchemaObject.items.type);
+                                    props.push(arrayProperty);
                                 }
                                 else {
                                     const objectProperty = new TypeDefinitionObjectProperty(propertyName + "[]", propertySchemaObject.items, isRequired, true);
                                     props.push(objectProperty);
                                 }
 
-                                props.push(arrayProperty);
                                 break;
 
                             case "combination":
--- a/src/models/typeDefinition.ts
+++ b/src/models/typeDefinition.ts
@@ -245,7 +245,7 @@ export class TypeDefinitionObjectProperty extends TypeDefinitionProperty {
                             case "object":
                                 const objectProperty = new TypeDefinitionObjectProperty(propertyName, propertySchemaObject, isRequired, true);
 
-                                if (!propertySchemaObject.$ref && !nested) {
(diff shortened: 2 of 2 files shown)


frankdewaelle:
When will this fix be deployed?

mikebudzynski:
The PR was merged over 3 months ago and it was deployed shortly after merging. Make sure to republish your portal.

frankdewaelle:
Tx for the info.
How can I check the version my portal is built on?

mvg,
Frank.

mikebudzynski:
There is no way to check the portal's version, unless you're using the self-hosted portal.

frankdewaelle:
Ok, thanks.
mvg,
Frank.

Pcaito:
im still having the issue with the object/array showing the elements properly.  republished the portal and still have the issue.

azaslonov:
Hi @Pcaito, can you please provide a snippet of your specification and roughly describe how you expect to see it rendered (just to be on the same page)? We've done bunch of improvements recently, but not everything got deployed yet.

Pcaito:
this is an example of the request:
[
    {
        "ActionCode": "string",
        "ScheduleNumber": 0,
        "Description": "string",
        "Type": 0,
        "DaysOfWeek": [
            0
        ],
        "DeliveryTime": "string",
        "SubmitByDay": 0,
        "SubmitByTime": "string",
        "Supplier": 0,
        "Stores": [
            0
        ],
        "SupplierItems": [
            "string"
        ]
    }
]

but when published to the portal it shows as the below with no details on the fields.

request

azaslonov:
Yeah, looks like the one we fixed in ongoing release. Please try to check again by the end of next week.

wolfgangas:
I noticed that this fix resulted in substantial improvment, now description is shown for every field. But I get [] (opened/closed square brackets) in front of every field name. I suppose that is not intended. It seems to be shown for every field that is member of an array. 

Here is my definition (initial part):
```
{
    "type": "array",
    "items": {
        "type": "object",
        "properties": {
            "kampagneId": {
                "type": "integer",
                "description": "campaign id"
            },
            "bezeichnung": {
                "type": "string",
                "description": "name of the campaign"
            },
            "von": {
                "type": "string",
                "description": "start date of the campaign"
            },
            "bis": {
                "type": "string",
                "description": "end date of the campaign"
            },
            "zuordnungen": {
                "type": "array",
                "items": {
                    "type": "object",
                    "properties": {
                        "kampagneZuordnungId": {
                            "type": "integer",
                            "description": "id for assignment of offer to campaign"
                        },
```

and the result in the developer portal table:
```
Name | Required | Type | Description
--- | --- | --- | ---
[].kampagneId  | true |  integer | campaign id
[].bezeichnung  | true | string | name of the campaign
[].von | true | string | start date of the campaign
[].bis | true | string | end date of the campaign
[].zuordnungen[].kampagneZuordnungId | true | integer | id for assignment of offer to campaign
and so on...
```

mikebudzynski:
@wolfgangas, `[]` indicates an array. In the example above, `[].kampagneId` would mean that the custom object that the array's elements represent has a property `kampagneId`. If you'd rather avoid `[]`, you can define object schema explicitly, instead of specifying it inline.

wolfgangas:
@mikebudzynski thanks for your advice. As can be seen from my definition above, I have a schema with an array of objects (campaigns / kampagne). These objects contain string and integer attributes, but also another array of objects (assignments / zuordnungen). So I did the following: within the Definitions of the API in Azure Portal I created a definition "zuordnung". Then I created the "parent" definition "kampagne" which refers to the subordinate definition like this (I removed the other parts): 
```
"zuordnungen": {
	"type": "array",
	"items": {
		"$ref": "#/components/schemas/zuordnung"
	}
}
```
On top of this I changed my original definition to this
```
{
    "type": "array",
    "items": {
		"$ref": "#/components/schemas/kampagne"
    }
}
```
I noticed that the validator of the definition dialogue does not check whether the reference exists, but this is just a side note. The result within the Developer Portal is as follows:

image
As you can see in this screenshot the top section of the response shows just a single line which is the definition on top of the other 2. 

I see all 3 definitions further below in a separate section "Definitions". I think this is way harder to comprehend for someone trying to understand the data model of the response. So if there are no other options I will switch back to "single view" with lots of []. 

I understand that somehow the information of objects being part of an array needs to be communicated, but I think that should be possible, have a look at this example [Clover Docs](https://docs.clover.com/reference/merchantcreatemerchantopeninghours). Array of objects are clearly shown in this documentation.

PR #849, opened 2020-08-25
Added files for backing up and restoring the portal

### LINKED ISSUE #848: Scripts for easily backing up and restoring the portal
## Bug description

When customizing the portal it is impossible to undo a change once you have saved it. That makes it risky to work with it. There are two scripts that exist that let you save the customizations; migrate.js and capture.bat. However, capture.bat does not save the media files and migrate.js uses hard coded paths and requires information about a destination portal. Therefore it would be great if there was a script for easily backing up the customizations, preferably with an option to add a timestamp to the file name and the folder name.

If you are able to specify the paths you would be able to put the files directly in a folder where it may be connected to a repository.

Also, an accompanying script for restoring/bringing up a portal using the saved files would be great. That would enable it to work with continuous delivery.

ErikMogensen:
This would fix #848 

Most of the code is copied from migrate.js. To remove the duplicated code, migrate.js could use the functions in shared.js.

### DIFF
--- a/scripts/backup.js
+++ b/scripts/backup.js
@@ -1,0 +1,169 @@
+"use strict"
+/**
+ * This script saves the developer portal instance.
+ * In order to run it, you need to:
+ * 
+ * 1) Clone the api-management-developer-portal repository
+ * 2) `npm install` in the root of the project
+ * 3) Install az-cli (https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest)
+ * 4) Run this script with a valid combination of arguments
+ * 
+ * Managed portal command example:
+ * node backup --sourceEndpoint from.management.azure-api.net --publishEndpoint to.developer.azure-api.net --sourceToken "SharedAccessSignature integration&2020..."  --dataFile d:\apim-portal\great.json --mediaFolder d:\apim-portal\great\
+ * 
+ * If you run with the --selfHosted flag, you are expected to supply a sourceStorage parameter.
+ * 
+ * You can specify the SAS tokens directly (via sourceToken), or you can supply an identifier and key,
+ * and the script will generate tokens that expire in 1 hour. (via sourceId, sourceKey).
+ */
+
+const path = require("path");
+const mkdirSync = require('fs').mkdirSync;
+const execSync = require('child_process').execSync;
+
+const shared = require('./shared');
+
+const yargs = require('yargs')
+.version('1.0.2')
+    .example('$0 \
+        --sourceEndpoint contoso.management.azure-api.net \
+        --sourceToken <token> \
+        --dataFile d:\apim-portal\contoso.json \
+        --mediaFolder d:\apim-portal\contoso \n')
+    .example('$0 \
+        --sourceEndpoint great.management.azure-api.net \
+        --sourceId integration \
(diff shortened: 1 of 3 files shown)


shashiguru:
Hi there, i have reviewed the code of 3 files but as we already have .bat files for generate, capture. we can use those to do backup and restore. do we really need these files now?

ErikMogensen:
Hi,

Since capture.bat does not save the media files my initial plan was to just create a bat file that called capture.bat and then did the az storage ... command but I changed my mind and made a Javascript wrapping those calls for the following reasons:

- Need to create the folder for the media files
- Handling the complex parameters, yargs does an excellent job of doing that
- Ability to save often without overwriting the previous save

ErikMogensen:
Another way of solving the main issue would be to add the possibility to save media files to capture.bat and capture.js and the upload functionality to the generate.* files.

 It would be great if the ability to save often without overwriting the previous save was added to the capture.* files too.

azaslonov:
Hi folks, just wanted to update you. We need to land an important change around API and related scripts before merging this PR. When done, we'll give specific pointers of what needs to be changed (if anything). Thanks for your contribution.

azaslonov:
Hi @Erik, thanks for your contribution, we greatly appreciate your help!

However, I think @shashiguru is right - there are files doing similar things. I was thinking to re-work capture/generate JS files into what you have in backup/restore JS files and just change the invocation of those files in capture/generate BAT files. This way we could probably solve both concerns. Also, I think naming "backup" and "restore" is more appropriate as well.

Therefore, maybe if you have spare time and willing to re-work these scripts a bit, we'd be even more grateful. If not, we'll do it in the next couple of weeks.

ErikMogensen:
Hi Alex,

I agree changing the capture/generate JS files is better. It requires more work though since you would have to change depending scripts, but I see that you already have done these changes to the scripts in scripts.v2 :smiley:. I also see that you got rid of the dependency to Azure CLI, which is great 👍 

As far as I can tell there are some features that are not there; the ability to publish when running generate.js and the ability to add a timestamp when saving. I also liked the functionality of yargs for the parameter handling.

I will happily add those features, but do you want them?

azaslonov:
> Hi Alex,
> 
> I agree changing the capture/generate JS files is better. It requires more work though since you would have to change depending scripts, but I see that you already have done these changes to the scripts in scripts.v2 😃. I also see that you got rid of the dependency to Azure CLI, which is great 👍
> 
> As far as I can tell there are some features that are not there; the ability to publish when running generate.js and the ability to add a timestamp when saving. I also liked the functionality of yargs for the parameter handling.
> 
> I will happily add those features, but do you want them?

Hi Erik, sorry for the late response. Yes, actually if we could have those (timestamp and publishing flag) as parameters, I think it would be awesome.

ErikMogensen:
No problem. Glad to see that you want those features. I'll add them during the next month.

jake-scott:
Hi.. I actually ported the JS files to a Golang CLI that can be used to backup/restore/reset the portal :+1: 

  https://github.com/jake-scott/apim-tools

The JS tools were a useful reference, but they are not production ready, and require the user present a SAS token.  The tools above obtain the SAS token from the Azure RM API, and authenticate the caller using standard Azure AAD.

I hope that is useful!

azaslonov:
Closing PR, since the mentioned logic was implemented in #1031.

@jake-scott, to my shame I noticed your comment just now, your work probably would have saved me a lot of time. I see your solution also based on ARM and `az login` approach, plus that you've done some nice reporting and more. I hope you don't mind us stealing some things :)

PR #876, opened 2020-09-03
Made blob storage connection string and SAS URL optional.

azaslonov:


ygrik (approved):

velsietis:
Bumping `@paperbits/azure` to `0.1.317` breaks image links in the Designer tool.

I believe it's because SharedKeyCredential doesn't work in the browser so the [if check](https://github.com/paperbits/paperbits-azure/commit/9970dfe6593accd0a9b38f0dcdacdb8d3f3b5661) errors.

That error causes it to fall back to the cached storage urls (which in my case have expired SAS tokens)

PR #893, opened 2020-09-07
Added missing await in migrate.js script.

mathiasi:
Missing await


azaslonov (approved):
Thanks @mathiasi, that's a good finding.

PR #990, opened 2020-10-21
Add Content-Type to `uploadBlobs` function

TomBonnerAtTFL:
Uses the package `mime-types` to calculate the file extension from the content-type when it downloads blobs, saves that with the extension, then uses the extension to calculate a content-type to upload with.


azaslonov (approved):
Hi Tom, that's a good one! Thanks a lot for your contribution.

mikebudzynski:
Thanks for your contribution @TomBonnerAtTFL!

Issue #998, opened 2020-10-28
Migrate errors - UnhandledPromiseRejectionWarning  and SAS Token expiry, etc


TimMcCann:
We have been spending considerable time troubleshooting the migrate process for self-hosted Dev Portal. Most of the issues are centered around 1) Debugging "UnhandledPromiseRejectionWarning" and 2) using SAS tokens that are expired. 
Not enough information is being logged. In fact, the logic is missing a catch that would benefit error handling. 

And there are similiar Github issues for this context. [458](https://github.com/Azure/api-management-developer-portal/issues/458) , [585](https://github.com/Azure/api-management-developer-portal/issues/585) 

The fix to overcoming these issues is to add a "catch" and also "reject" non 200 status codes.
For example migrate.js  calls 
	• capture.js
	• generate.js 
	• cleanup.js 
  
Below are the changes to **capture.js** to log errors   

**Add a catch and log the error**
![image](https://user-images.githubusercontent.com/38544546/97480360-85ce6800-1910-11eb-8a60-cf2ace1eef1e.png)
   
**Reject and log a 400 status**
![image](https://user-images.githubusercontent.com/38544546/97480633-eb225900-1910-11eb-8842-1d5349193962.png)

**Summary**
Adding the "catch" and "reject" logic has surfaced errors that have been buried.  The changes can be validated by running capture.bat and trying good and bad values for the environment variables passed in.  It would be nice to see these changes make it into the base code.

BobbyCGD:
Why not open a pull request?

azaslonov:
Hi @TimMcCann, I feel your pain, and it makes total sense. We'll fix this shortly, unless you make it first as @BobbyCGD suggested.

TimMcCann:
@BobbyCGD @azaslonov Thanks, for calling that out. I will do a pull request. Wanted to include generate.js also. But didn't want to step on my teams current work. This is all a little confusing ... V1 no longer supported.....and migrate documentation says Self-hosted use V1.

mikebudzynski:
@TimMcCann - it depends on the configuration of the self-hosted portal. If your portal relies on the storage account built-into API Management, you should use the latest scripts. If your portal uses an external, explicitly-defined storage account, you need to use the v1 scripts.

azaslonov:
Closing the issue, since the pull request has been merged. @TimMcCann, thanks again!

PR #1011, opened 2020-10-31
Updated capture.js (v1) to output errors in case of failure.

TimMcCann:
Per Issue 998. https://github.com/Azure/api-management-developer-portal/issues/998
Add a catch and check for 200 response code in scripts/capture.js

### DIFF
--- a/scripts/capture.js
+++ b/scripts/capture.js
@@ -1,7 +1,3 @@
-/*
- * Important: this script is no longer maintained; new scripts are located in the scripts.v2 folder.
-*/
-
 const fs = require("fs");
 const https = require("https");
 const managementEndpoint = process.argv[2];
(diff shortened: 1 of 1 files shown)


scripts/capture.js
```diff
@@ -22,18 +18,21 @@ async function request(url) {
...
+				// reject on bad status
+				if (resp.statusCode != 200) {
+				  console.log('url: ' + url);	
+                  return reject(new Error('statusCode=' + resp.statusCode + ' for URL: ' + url));
+                }
                 try {
                     resolve(JSON.parse(data));
                 }
-                catch (e) {
+                catch (e) {;
```
azaslonov, line 33: Do you mind removing the unnecessary `;` sign here? Also, it would be great if you could use `Shift+Alt+F` to apply formatting on the document.

azaslonov (changes requested):
Thank you for your contribution!

azaslonov (approved):
Thanks a lot!

PR #1044, opened 2020-11-20
Fix Operation Console for SOAP 1.2 Content-Type

velsietis:
The "Try It" console doesn't display the correct headers for SOAP 1.2 APIs due to some minor errors in the code:

 - `ConsoleHeader.value` has tyoe `ko.Observable<string>` so needs to be called as function
 - `consoleOperation.setHeader()` pushes a new Content-Type header, so instead just update the value of the existing one


azaslonov (approved):
Hi @velsietis, that's a nice one! Thanks for your contribution.

PR #1080, opened 2020-12-09
Make css change to support user-guide html table formatting

bmcd77:


bmcd77:
Nope this is not what I was after! Closing this pull request.

PR #1137, opened 2021-01-29
Added basic XSD schema support in API documentation.

azaslonov:
Added basic XSD schema support in API documentation. Now examples and schemas for XML payloads get displayed in similar manner as JSON payloads.
image


src/models/jObject.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    public getAttributeAsNumber(attributeName: string): number {
+        const value = this.getAttribute(attributeName);
+
+        try {
+            return parseInt(value);
+        }
+        catch (error) {
+            return undefined;
+        }
```
ygrik, line 254: this block can be simplified as
return +value || undefined;

azaslonov: Done

ygrik (approved):

Issue #1144, opened 2021-02-09
Migrate command using script v3 fails for migration of apim developer portal from one subscription to another subscription


SSanjeevi:
## Bug description

Migrate command using script v3 fails for migration of apim developer portal from one subscription to another subscription. Apim 1 in one subscription A and apim 2 is in subscription B when using script v3 for migrating the portal it fails with following error:

"Unable to complete migration. Unable to complete cleanup. Unable to delete content. Unable to fetch content types. Unauthorized. Make sure you're logged-in with "az login" command before running the script."

## Reproduction steps

1. Create apim 1 in A subscription and apim2 in B subscription
2. use migrate v3 command to migrate from apim1 to apim 2.
3. error at cleaning up "Unable to complete migration. Unable to complete cleanup. Unable to delete content. Unable to fetch content types. Unauthorized. Make sure you're logged-in with "az login" command before running the script."

## Expected behavior

Apim developer portal import should complete.

## Is your portal managed or self-hosted?

both are Managed portals

## Release tag or commit SHA (if using self-hosted version)

Latest version.

## API Management service name

will share in email.

## Environment

NA

## Additional context

Am executing the command in Azure devops task.
Tried using Azure cli task, powershell task it works for apim in same subscription and tenant but fails for different subscription.

<img width="699" alt="apim dev portal migrate command v3 issue" src="https://user-images.githubusercontent.com/24915474/107345414-88cd1d00-6ae9-11eb-9486-a53cae078d02.PNG">

SSanjeevi:
I managed to succeed by using this PR changes.

Where i have added code to login to az cli with each subscription on source and destination.

https://github.com/Azure/api-management-developer-portal/pull/1145

SSanjeevi:
I managed to succeed by  using this PR changes 
https://github.com/Azure/api-management-developer-portal/pull/1145

hammdo:
What permissions are necessary for the service principal in order to perform a clean/delete and import?  Currently, we have:

Register app to create service principal, create secret

add role assignment with APIM READ access DELETE and WRITE access on the app

add the service principal as the contributor in APIM

We're failing with:

Cleaning up...
node : Unable to complete migration. Unable to complete cleanup. Unable to delete content. Unable to fetch content types. Looks like you are not allowed to perform this operation. 
Please check with your administrator.
At F:\projects\Rogue-One\dev-portal\api-management-developer-portal\scripts.v3\DevPortalMigration.ps1:144 char:7
[...]

PR #1145, opened 2021-02-09
Fix for migration between subscription-to-subscription apim developer portal

SSanjeevi:
### Fix for migration between subscription-to-subscription apim developer portal by passing the az login details in parameters

Migration failing if the source and destination is on different subscription because azure cli access is not there in one apim which causes unauthorized error on generating sas token to migrate.

So parameterized the az cli login details to the command line for source and destination from command arguments below:

Following are the parameters added to pass while running and are optional for users

 *    --sourceTenantid "< your service name >" ^
 *    --sourceServiceprincipal "< source serviceprincipal or user name. >" ^
 *    --sourceSecret "< secret or password for service principal or az login for the source apim. >" ^
 *    --destTenantid "<  destination tenantid >"
 *    --destServiceprincipal "<destination serviceprincipal or user name. >"
 *    --destSecret "< secret or password for service principal or az login for the destination. >"

its a fix for issue : https://github.com/Azure/api-management-developer-portal/issues/1144

### DIFF
--- a/scripts.v3/migrate.js
+++ b/scripts.v3/migrate.js
@@ -13,15 +13,15 @@
  *    --sourceSubscriptionId "< your subscription ID >" ^
  *    --sourceResourceGroupName "< your resource group name >" ^
  *    --sourceServiceName "< your service name >" ^
- *    --sourceTenantid "< source tenant id >" ^
- *    --sourceServiceprincipal "< source serviceprincipal or user name. >" ^
- *    --sourceSecret "< secret or password for service principal or az login for the source apim. >" ^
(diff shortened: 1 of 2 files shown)


azaslonov:
Hi @SSanjeevi, thanks for your contribution. This is a nice addition, we didn't think of this scenario. The only thing that would be great to add is an indication that those parameters are optional, and needed for this specific case.

SSanjeevi:
@azaslonov - I have updated the description, please verify it.

azaslonov (approved):

azaslonov:
Thanks again Sanjeevi! We appreciate your contribution.

PR #1149, opened 2021-02-16
Process for migration scripts will return when underlying script fails

marcin-dudek:
Currently during migration run errors that happen in underlying scripts (executed within execSync) got swallowed.

After the change, execSync will trigger exception when non-zero return code is received and you will receive error to console.


azaslonov (approved):
Thanks Marcin, makes sense, thank you for your contribution!

PR #1159, opened 2021-02-22
Changed variable requestUrl to url in utils.js

ErikMogensen:
The variable requestUrl which is used in two places in the request method in utils.js is never assigned. It looked like url was the variable intended to be used.


azaslonov (approved):
Thanks Erik, makes sense.

PR #1164, opened 2021-02-24
Fix for parameter check in migrate.js

ErikMogensen:
It should check for destId or **destKey** when the destToken parameter is being used. It didn't check for destKey.


azaslonov (approved):
Thanks Erik!

PR #1189, opened 2021-03-16
Added support for tags as a reference in URL for apis and operations

### LINKED ISSUE #1073: Pass tags to filter by as URL parameters for the list of APIs and list of API operations widgets
It would enable creating hierarchies of APIs through static links.

ygrik:
close #1073

### DIFF
--- a/src/components/apis/list-of-apis/ko/runtime/api-list.ts
+++ b/src/components/apis/list-of-apis/ko/runtime/api-list.ts
@@ -8,7 +8,6 @@ import { TagGroup } from "../../../../../models/tagGroup";
 import { SearchQuery } from "../../../../../contracts/searchQuery";
 import { RouteHelper } from "../../../../../routing/routeHelper";
 import { Tag } from "../../../../../models/tag";
-import { Utils } from "../../../../../utils";
 
 
 @RuntimeComponent({
@@ -61,6 +60,7 @@ export class ApiList {
     @OnMounted()
     public async initialize(): Promise<void> {
         this.groupByTag(this.defaultGroupByTagToEnabled());
+        this.tags.subscribe(this.resetSearch);
 
         await this.resetSearch();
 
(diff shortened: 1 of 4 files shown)


azaslonov (approved):

AnRei123:
Not sure how does this works. Could you make an example how to use this option?

mikebudzynski:
Filtering by tags is now passed in the URL parameter. For example: https://contoso.developer.azure-api.net/apis#tags=testtag would navigate to the APIs page and automatically filter the APIs list by the "testtag" tag.

ncroad:
Are we able to confirm that this functionality is still working as expected? If the appropriate thing is to raise a defect or issue, please let me know.
When you add a filter manually on the api list page, it works fine. If you refresh the page, or navigate to the page with the tag filter it doesn't filter the list 100% of the time, actually it is very hard to get it to work.

We had this previously working late last year, early this year, but it appears now that it is intermittent and in same cases the second API call that includes the tags as a query string parameter doesn't always fire. 
Where we have seen it working the 2nd api call with tags, has only been when it completes after the first call for the list of apis. If it completes before the one with the tags it appears to show the full list.  We are using the noted URL format above. 
API calls I am referring to from the portal are: 
https://xxxxxxxxxxx/subscriptions/000/resourceGroups/000/providers/Microsoft.ApiManagement/service/apim-instance-xxxxx/apis?expandApiVersionSet=true&$top=50&$skip=0&$filter=isCurrent%20eq%20true&api-version=2021-04-01-preview
vs: 
https://xxxxxxxxxxx/subscriptions/000/resourceGroups/000/providers/Microsoft.ApiManagement/service/apim-instance-xxxxx/apis?expandApiVersionSet=true&$top=50&$skip=0&tags[0]=TagXXXXX&$filter=isCurrent%20eq%20true&api-version=2021-04-01-preview

PR #1197, opened 2021-03-23
Fixed some issues with hyperlink and anchor resolution.

azaslonov:


ygrik (approved):

alilnik:
@azaslonov good time of the day!

Sorry for the question out of nowhere, but we recently had problems with Sign Out of the Developer Portal (with delegation). Today I found that Sign Out is working as expected. MS Azure support was also surprised by that. Do you think this PR may have affected Sign Out logic in Developer Portal? 

Best,
I

PR #1215, opened 2021-03-29
Increase apiversion to handle the case when minimum apiversion has been set

ErikMogensen:
If you set "Prevent users with read-only permissions from accessing service secrets" it will set the Minimum API version to 2019-12-01 for the Management API.

That stops the capture and generate scripts in the scripts.v3 folder from working. 

This PR increases the apiVersion used in the scripts from 2019-01-01 to 2020-06-01-preview, which makes them work again.


azaslonov (approved):
Makes sense, thank you!

PR #1233, opened 2021-04-06
Extended "Operation details" widget with "Use CORS proxy" option.

azaslonov:


ygrik (approved):

AnRei123:
Have I understood correctly that the behavior of the Try it function is for the following cases the same?
- "Use CORS proxy" option in the API operation details widget 
- Use a global policy to enable the CORS proxy
Have I understood correctly that in both cases, the interactive console's API calls are rooted through the portal's backend in my API Management service?
So why have you introduced the function in the operation details widget? To get get of the related policy entries? Or was there also another reason? 
I feel a little bit lost with this new function? What is your recommendation? In which cases is it better to add activate the CORS proxy in the policy and in which case would it be better to just activate the "Use CORS proxy" settings? 
Is it a problem if the policy states to use the CORS proxy and the "Use CORS proxy" hasn´t been activated in the operation details widget? Would the call be rooted nevertheless to the portal´s backend in my API Management service?

azaslonov:
> Have I understood correctly that in both cases, the interactive console's API calls are rooted through the portal's backend in my API Management service?

Yes, with "Use CORS proxy" option enable the requests from Try console will go through the portal's backend.

> So why have you introduced the function in the operation details widget? To get get of the related policy entries? Or was there also another reason?

There are a bunch of scenarios related to VNet (i.e. additional network configuration), AppGateway (i.e. false positives in "prevention" mode), and some browser limitations (i.e. override certain headers like "Origin") that we needed to address. Plus that, the CORS proxy allows not to configure CORS policies.

> Is it a problem if the policy states to use the CORS proxy and the "Use CORS proxy" hasn´t been activated in the operation details widget?

Nope, not a problem.

> Would the call be rooted nevertheless to the portal´s backend in my API Management service?

Only if "Use CORS proxy" option is enabled.

AnRei123:
For our test instance, I observed the following behavior. After I activated the "Use CORS proxy" option in the API details widget, when using the Try it panel to test an API, no response has been received anymore. Is there any additional setting in the APIM portal required to support this option for the managed approach? Or why does it not work for me?

mikebudzynski:
> For our test instance, I observed the following behavior. After I activated the "Use CORS proxy" option in the API details widget, when using the Try it panel to test an API, no response has been received anymore. Is there any additional setting in the APIM portal required to support this option for the managed approach? Or why does it not work for me?

It's working for me. Is your API accessible from the API Management service? Is it deployed in a virtual network / self-hosted API gateway?

PR #1279, opened 2021-04-27
use update instead of init in markdown bindingHandler

patst:
I followed the guide to implement a custom widget (https://docs.microsoft.com/en-gb/azure/api-management/developer-portal-implement-widgets ) but had problems getting the markdown binding working.

After some debugging I found out it seems to be related to the `httpClient.send` await.
The markdown binding is called before the http response is returned with an empty value.

After taking a look at the other custom `bindingHandlers` they all use the `update` callback method (see https://knockoutjs.com/documentation/custom-bindings.html ).


azaslonov (approved):
Thanks for your contribution!

PR #1346, opened 2021-06-17
Leveraged B2C signout. Moved AAD/B2C configuration publishing into config.json. Added option to override reply URL(s).

### LINKED ISSUE #1068: Developer Portal to set up Sign Out with OAUTH
**Insight:** We enabled developer portal user sign in with OAUTH.
Azure active directory is our IDP & this is updated in identities in APIM.
Developer portal is registered as client in AAD.

We are using Sign in with OAUTH function.
The sign in works as expected with OAUTH.

``
**Issue** is with sign out process.
There is no option to Sign out with OAUTH currently provisioned. 

Hence we are using standard Signout function.

This is only enabling our users to sign out from their Developer portal session.
This does not allow AAD session to be logged out.
SO when user signs back in, it does not redirect them to go through login process, it just signs right back in.
This is a security vulnerability & does not comply with our security requirements for 2FA.

We raised case with Microsoft.
After reviewing our request, Microsoft Product Group recommended to request as open issue on GITHUB.

azaslonov:
Closes #1068.


ygrik (approved):

wkarratoon:
Please explain in a sequenced way what a Developer Portal developer using a B2C tenant is expected to do to get *full* sign-out (both B2C and default Developer Portal SAS token) working. I have tried various reply url settings in the modified B2C sign-in button and likewise for my custom sign out button and *nothing* works for complete sign out. Same issue is present. 
I was expecting that a simple sign out button link to /#signout or /signout and this would take care of everything - detecting you are using a B2C tenant as configured in the administrative interface. (There is no update to the documentation I could find either.)

PR #1365, opened 2021-07-07
Support for setting folders and timestamp for the capture and generate scripts

### LINKED ISSUE #848: Scripts for easily backing up and restoring the portal

ErikMogensen:
Currently the capture and generate scripts are hardcoded to work with folders that are relative to the current working directory  which makes it difficult to determine where portal content is stored and picked up from. It also limits the options where content may be stored.

This PR enables setting the folder paths. It also enables saving the content of the portal in a new directory every time using a --timestamp argument.

This fixes #848.

It has only been test on Windows.

### DIFF
--- a/scripts.v3/capture.js
+++ b/scripts.v3/capture.js
@@ -10,21 +10,19 @@
  * 
  * 3) Run this script with a valid combination of arguments:
  *    node ./capture ^
- *    --sourceSubscriptionId "< your subscription ID >" ^
- *    --sourceResourceGroupName "< your resource group name >" ^
- *    --sourceServiceName "< your service name >" ^
- *    --destSubscriptionId "< your subscription ID >" ^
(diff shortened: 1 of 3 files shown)


azaslonov (approved):
Thanks for contribution Erik!

PR #1396, opened 2021-07-21
fix: capitalization of filename

BernhardRode:
filename is not capitalized correctly, which leads to build failures on linux systems (docker node:lts).


BernhardRode:
There is already a pr available =)

PR #1401, opened 2021-07-22
Made the update existing URLs properties optional so that it doesn't breaks any existing code

SSanjeevi:
### Made the update existing URLs properties optional so that it doesn't breaks any existing code


Pipelines/Pipeline.yml
```diff
@@ -0,0 +1,28 @@
...
+      name: $(poolName)
+    timeoutInMinutes: 90
+    steps:
+      - task: Npm@1
+        displayName: Npm Install command
+        inputs:
+          command: "install"
+
+      - pwsh: |        
+          node ./migrate --sourceSubscriptionId "$(sourceSubscriptionId)" --sourceResourceGroupName  "$(sourceResourceGroupName)" --sourceServiceName "$(sourceAPIMName)" --destServiceName "$(destinationAPIMName)" --destSubscriptionId "$(destSubscriptionId)" --destResourceGroupName "$(destResourceGroupName)" --sourceTenantid "$(sourceAzure_Tenant)" --sourceServiceprincipal "$(sourceServicePrincipal)" --sourceSecret "$(sourceAzureDevOps-ServicePrincipal-Secret)"  --destTenantid "$(Azure_Tenant)" --destServiceprincipal "$(ServicePrincipal)" --destSecret "$(AzureDevOps-ServicePrincipal-Secret)" --existingEnvUrls "$(existingEnvUrls)" --destEnvUrls  "$(destEnvUrls)"
```
azaslonov, line 26: I'm not expert, but I think these variable should be declared in variables section, like [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#macro-syntax-variables) suggests. Example:
```
variables:
 - name: projectName
   value: contoso

steps: 
- bash: echo $(projectName)
- powershell: echo $(projectName)
- script: echo $(projectName)
```

Pipelines/Pipeline.yml
```diff
@@ -0,0 +1,28 @@
+name: CD-ApiMDeveloperPortal-Build
```
azaslonov, line 1: There is common practice to put all pipeline files into `.pipeline` folder. Also, I would give the file specific name, i.e. `migrate.yml`.

azaslonov (changes requested):
@SSanjeevi, are you sure you want URL replacement to be part of migration script? This looks like very narrow and specific case that maybe not be needed by anyone else (though I would love to hear feedback from community). Because, if that's the case, it just adds more complexity to the usage. Probably, it would make sense to put it into separate script designated to content updates/adjustments. Thoughts?

scripts.v3/utils.js
```diff
@@ -418,13 +406,43 @@ class ImporterExporter {
...
+            for (const contentItem of contentItems) {
+                var count = 0;
+                console.log("url found in portal: " + contentItem.properties.permalink);
+
+                for (const existingUrl of existingUrls) {
+                    if (contentItem.properties.permalink == existingUrl) {
+                        contentItem.properties.permalink = replaceableUrls[count];
+                        console.log("updating url content..." + contentItem.properties.permalink);
+                        console.log(" updated url content : " + JSON.stringify(contentItem));
+                        const response = await this.httpClient.sendRequest("PUT", contentItem.id + "?api-version=2019-12-01", contentItem);
```
azaslonov, line 432: Please use `apiVersion` constant defined above in this file rather the in-place version.

scripts.v3/utils.js
```diff
@@ -418,13 +406,43 @@ class ImporterExporter {
...
-        catch (error) {
+        } catch (error) {
             throw new Error(`Unable to schedule website publishing. ${error.message}`);
         }
     }
+
+    /**
+     * Updates the url links of specified API Management service into updated url in destination.
+     */
+    async updateContentUrl(existingUrls, replaceableUrls) {
```
azaslonov, line 417: `async updateContentUrl(existingUrls, replaceableUrls)`
Since it depends on matching the number of exiting and replaceable URLs, it would be great to have a check for that and respective error.

Pipelines/Pipeline.yml
```diff
@@ -0,0 +1,28 @@
...
+variables:
+  - name: poolName
+    value: "agentpoolName"
+  - name: location
+    value: West US
+
+jobs:
+
+  # All tasks on APIM Developer portal pipeline
+  - job: Deploy_APIM_Developer_Portal
```
azaslonov, line 14: `- job: Deploy_APIM_Developer_Portal`
I assume, it's not a deployment, it is migration. Could you please adjust the wording here and everywhere in the file?

azaslonov (changes requested):

Pipelines/Pipeline.yml
```diff
@@ -0,0 +1,28 @@
+name: CD-ApiMDeveloperPortal-Build
+
+trigger: none
+
+variables:
+  - name: poolName
+    value: "agentpoolName"
+  - name: location
+    value: West US
```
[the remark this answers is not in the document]
SSanjeevi: removed it.

Pipelines/Pipeline.yml
SSanjeevi: Added readme file with details of pipeline variables to be added with value descriptions.

scripts.v3/utils.js
SSanjeevi: Created a variable.

scripts.v3/utils.js
SSanjeevi: Added a validation with description

Pipelines/Pipeline.yml
SSanjeevi: Updated all places.

Pipelines/Pipeline.yml
SSanjeevi: updated.

SSanjeevi:
@azaslonov we can move this to separate script file, but then i must duplicate the util js script file where i use many methods from this.
Am weak at NodeJS scripting that's why tried to reuse in single script.

SSanjeevi:
@azaslonov - please review and update your comments if you have anything.

scripts.v3/utils.js
```diff
@@ -6,6 +6,7 @@ const { BlobServiceClient } = require("@azure/storage-blob");
 const blobStorageContainer = "content";
 const mime = require("mime");
 const apiVersion = "2020-06-01-preview"; // "2021-01-01-preview"; 
+const apiVersion2019 = "2019-12-01";
```
azaslonov, line 9:
```
const apiVersion = "2020-06-01-preview"; // "2021-01-01-preview"; 
const apiVersion2019 = "2019-12-01";`
```

@SSanjeevi, please let's not use different API versions in the script, this pattern may lead to adverse results in the long term. Using `apiVersion` should be enough.

scripts.v3/utils.js
```diff
@@ -418,13 +407,48 @@ class ImporterExporter {
...
             throw new Error(`Unable to schedule website publishing. ${error.message}`);
         }
     }
+
+    /**
+     * Updates the url links of specified API Management service into updated url in destination.
+     */
+    async updateContentUrl(existingUrls, replaceableUrls) {
+        try {
+            if (existingUrls.Count != replaceableUrls.Count) {
```
azaslonov, line 420: `existingUrls.Count` - property `Count` doesn't exist in standard JS arrays. What kind of type is expected here? Can we use `length` instead?

.pipeline/readme.md
```diff
@@ -0,0 +1,35 @@
+# Create your Pipeline
```
azaslonov, line 1: @mikebudzynski, I think you should review this Readme.md

scripts.v3/migrate.js
```diff
@@ -117,9 +131,13 @@ async function migrate() {
         const destIimporterExporter = new ImporterExporter(yargs.destSubscriptionId, yargs.destResourceGroupName, yargs.destServiceName, yargs.destTenantid, yargs.destServiceprincipal, yargs.destSecret);
         await destIimporterExporter.cleanup();
         await destIimporterExporter.import();
+
+        if (yargs.existingEnvUrls != "" && yargs.destEnvUrls != "") {
```
azaslonov, line 135: @SSanjeevi, I still think it should be part of separate a script file (and invoked separately) because it is uncommon feature and no customers ever expressed intent to do something like this. We can't accept project-specific changes in commonly used scripts, otherwise we end up with unnecessarily bloated `migrate.js`. Hope that makes sense.

azaslonov (changes requested):

scripts.v3/migrate.js
azaslonov: This PR is a good example for the pattern: #1405

scripts.v3/utils.js
SSanjeevi: Fixed it and tested it.

scripts.v3/utils.js
SSanjeevi: done

scripts.v3/migrate.js
SSanjeevi: Done as per your request added script file updatecontenturl.js file

azaslonov (approved):

azaslonov:
Thanks @SSanjeevi! I appreciate all the changes.

PR #1454, opened 2021-09-16
Accessibility: fixed tags widget keyboard trap

### LINKED ISSUE #1374: Accessibility, Section 508 Compliance: List of APIs widget keyboard trap
## Bug description

After opening the tag filter popup within the List of APIs widget, a keyboard trap is created in the popup elements that cannot be exited without navigating to another page or interacting via mouse.

## Reproduction steps

Using a List of APIs widget:
1.  Move keyboard focus to the Tag Search button on the widget using Tab.
2. Open the Tag Search popup using Spacebar or Enter.
3. The user is able to tab between the search icon and search field but unable to close the popup or move tab focus out of the popup without interacting with the mouse or navigating to another page.

## Expected behavior

The Tag Search popup is able to be entered and then exited with focus returning to the next logical place in the page's tab order after the button which opened the popup using only keyboard controls

Managed

## Environment

 - Operating system: MacOS 10.15.7 - Macbook Pro 15-inch, 2017 model
 - Browser: Google Chrome
 - Version: 91
 
## Additional context

Affected widget:
Keyboard-Trap

Related Section 508 test step: https://section508coordinators.github.io/TrustedTester/keyboard.html#check-212-no-keyboard-trap

ygrik:
fix #1374 Accessibility, Section 508 Compliance: List of APIs widget keyboard trap

### DIFF
--- a/src/components/tag-input/tag-input.html
+++ b/src/components/tag-input/tag-input.html
@@ -10,7 +10,7 @@
     <!-- /ko -->
     
     <div tabindex="0" class="tag-item tag-add" role="button" title="Add tags to search query"
-        data-bind="balloon: { template: { name: 'tagList', data: $component }, closeOn: $component.onDismiss }">
+        data-bind="balloon: { template: { name: 'tagList', data: $component }, closeOn: $component.onDismiss }, event: { keydown: $component.onTagBtnKeyDown }">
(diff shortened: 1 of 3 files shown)


azaslonov:
Fixed in scope of another item.

PR #1507, opened 2021-11-02
Masked out subscription keys #1436

### LINKED ISSUE #1436: Subscription keys should be masked out as it supposed to be a secret
Hi team,

We think that `Ocp-Apim-Subscription-Key` should not be visible in clear text in developer console UI as it suppose to be treated as a secret.

Steps to reproduce:
1- Navigate to "Try it" console for an API operation where you have an existing subscription.

Actual behaviour:
UI displays subscription key
image

Expected behaviour:
UI shouldn't show subscription key, more like what's in Azure Portal
image

malincrist:
Closes #1436


src/bindingHandlers/copyToClipboard.ts
```diff
@@ -1,10 +1,13 @@
 import * as ko from "knockout";
 
 ko.bindingHandlers["copyToClipboard"] = {
-    init: (element: HTMLElement, valueAccessor: () => string): void => {
-        const copyToClipboard = () => {
+    init: (element: HTMLElement, valueAccessor: () => (string | (() => Promise<string>))): void => {
+        const copyToClipboard = async () => {
```
azaslonov, line 5: If there is no `await` in the function, having `async` and `Promise` is not necessary in this case.

azaslonov (commented):
@malincrist, could you please attach the screenshot of how this looks like? Specifically, it would be great to see how the triggering happens.

src/bindingHandlers/copyToClipboard.ts
malincrist: There is an `await` for that promise that could be accepted as parameter. We need to accept it as Promise because `getPlainTextCodeSample` needs to await the `render` method from `TemplatingService`

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -247,9 +247,30 @@ <h3>Headers</h3>
...
                     <div class="input-group">
+                        <!-- ko if: !header.secret -->
                         <input type="text" autocomplete="off" class="form-control form-control-sm" placeholder="value"
                             spellcheck="false" aria-label="Header value"
                             data-bind="event: { keyup: $component.updateRequestSummary }, textInput: header.value">
+                        <!-- /ko -->
+                        <!-- ko if: header.secret -->
+                        <div style="width: 100%;">
+                            <!-- ko if: !header.revealed() -->
+                            <input type="password" autocomplete="off" class="form-control form-control-sm password-input"
```
azaslonov, line 258: `<input type="password" autocomplete="off"` - technically, you can re-use exiting markup and switch input type between "text" and "password" when toggling show/reveal. For that you can use `attr` binding. Example: `data-bind="attr: { type: myInputType }`.

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -247,9 +247,30 @@ <h3>Headers</h3>
                     <!-- /ko -->
                     <!-- ko if: !header.options || header.options.length === 0 -->
                     <div class="input-group">
+                        <!-- ko if: !header.secret -->
                         <input type="text" autocomplete="off" class="form-control form-control-sm" placeholder="value"
                             spellcheck="false" aria-label="Header value"
                             data-bind="event: { keyup: $component.updateRequestSummary }, textInput: header.value">
+                        <!-- /ko -->
+                        <!-- ko if: header.secret -->
+                        <div style="width: 100%;">
```
azaslonov, line 256: `<div style="width: 100%;">` - we should try to avoid placing styles into the markup and rely more on CSS, where we define style for every component. That allows us to change website/component appearance from one place. Otherwise, it's pretty hard to debug styling issues, especially on a larger scale.

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -247,9 +247,30 @@ <h3>Headers</h3>
...
+                            <!-- ko if: !header.revealed() -->
+                            <input type="password" autocomplete="off" class="form-control form-control-sm password-input"
+                                placeholder="value" spellcheck="false" aria-label="Header value"
+                                data-bind="event: { keyup: $component.updateRequestSummary }, textInput: header.value">
+                            <!-- /ko -->
+                            <!-- ko if: header.revealed() -->
+                            <input type="text" autocomplete="off" class="form-control form-control-sm password-input"
+                                placeholder="value" spellcheck="false" aria-label="Header value"
+                                data-bind="event: { keyup: $component.updateRequestSummary }, textInput: header.value">
```
azaslonov, line 265: `event: { keyup: $component.updateRequestSummary }` - instead `keyup` event you could [subscribe to changes](https://knockoutjs.com/documentation/observables.html#explicitly-subscribing-to-observables) of the observable: `myObservable.subscribe(() => { ... }`. That's because the field value may change not only on `keyup` - user can copy paste values with mouse or another script can update the value (for instance, we're setting up an OAuth token into `Authorization` header when user authenticates in Test console.

src/components/operations/operation-details/ko/runtime/operation-console.ts
```diff
@@ -749,7 +748,27 @@ export class OperationConsole {
     }
 
     public toggleRequestSummarySecrets(): void {
+        if (!this.secretsRevealed()) {
+            this.consoleOperation().request.meaningfulHeaders().forEach(header => header.revealed(true));
```
azaslonov, line 752: `this.consoleOperation().request.meaningfulHeaders().forEach(header => header.revealed(true))`
`this.consoleOperation().request.meaningfulHeaders().forEach(header => header.revealed(false))`

Maybe just this:
`this.consoleOperation().request.meaningfulHeaders().forEach(header => header.revealed(this.secretsRevealed()));`

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -247,9 +247,24 @@ <h3>Headers</h3>
                     <!-- /ko -->
                     <!-- ko if: !header.options || header.options.length === 0 -->
                     <div class="input-group">
+                        <!-- ko if: !header.secret -->
                         <input type="text" autocomplete="off" class="form-control form-control-sm" placeholder="value"
                             spellcheck="false" aria-label="Header value"
                             data-bind="event: { keyup: $component.updateRequestSummary }, textInput: header.value">
+                        <!-- /ko -->
+                        <!-- ko if: header.secret -->
+                        <div class="input-div">
```
azaslonov, line 256: Regarding `input-div` and `password-input` classes: it looks very similar to what `.input-group` was intended to do. We already have this class defined in forms.scss of website styles, but it is outdated a bit. Please take a look at [naming and structure](https://getbootstrap.com/docs/5.1/forms/input-group/) of Bootstrap for input groups, I would even steal those names from them. This would allow us to have common utility to complex inputs like your password-input, and at the same time we don't introduce a new CSS class for each such case.

You may have a question like "why don't we just use the whole Bootstrap library?" :) So, answering it in advance: This is because we have to generate component styles and their variations on the fly. However, in many ways we use similar approach to structuring and naming our generated classes, in which the Bootstrap team makes a very good job.

azaslonov (changes requested):

src/bindingHandlers/copyToClipboard.ts
malincrist: When the secrets are hidden, I replace each character of them with the character '•' in the `displayValue` property. When the user clicks copy, if he did not previously revealed the secrets, the `displayValue` will still be composed only from '•'. From what I understood, we don't want the user to copy the code template like this, but with the actual value. For this I have only found 2 solutions:
1. Keep two string rendered values one with the actual value and one with them replaced with '•' and always copy the one with plain values. But this means that we will have to render them both whenever a header changes.
2.  Only render the plain value when `copyToClipboard` is triggered and keep an already rendered value (depending on the `revealed` property) for the code sample.

Do you think the first option would be better? Or do you have another idea?

azaslonov: Ok, I see now. Looks like your original variant is the most optimal in this case. Let's keep it then.

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -305,20 +317,23 @@ <h3>Body</h3>
...
-            <span class="text-monospace item-label" data-bind="text: item.name"></span><span class="text-monospace" data-bind="text: '(type: ' + item.type() + ')'"></span>
+            <span class="text-monospace item-label" data-bind="text: item.name"></span><span class="text-monospace"
+                data-bind="text: '(type: ' + item.type() + ')'"></span>
         </div>
         <div class="form-group">
             <!-- ko if: item.bodyFormat() === 'string' -->
             <div class="input-group">
-                <input type="text" autocomplete="off" class="form-control form-control-sm" spellcheck="false" data-bind="event: { keyup: $component.updateRequestSummary }, textInput: item.body">
+                <input type="text" autocomplete="off" class="form-control form-control-sm" spellcheck="false"
+                    data-bind="event: { keyup: $component.updateRequestSummary }, textInput: item.body">
```
azaslonov, line 327: `event: { keyup: $component.updateRequestSummary }` - still got `keyup`. Please my previous comments.

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -247,9 +247,21 @@ <h3>Headers</h3>
...
                         <input type="text" autocomplete="off" class="form-control form-control-sm" placeholder="value"
                             spellcheck="false" aria-label="Header value"
                             data-bind="event: { keyup: $component.updateRequestSummary }, textInput: header.value">
+                        <!-- /ko -->
+                        <!-- ko if: header.secret -->
+                        <input autocomplete="off" class="form-control form-control-sm input-password"
+                            placeholder="value" spellcheck="false" aria-label="Header value"
+                            data-bind="attr: {type: header.inputTypeValue}, textInput: header.value">
+                        <button data-bind="click: () => $component.toggleSecretHeader(header)"
+                            class="toggle-secrets-button">
```
azaslonov, line 260: `class="toggle-secrets-button"` - this is still specific case naming (please see my previous comments). It could be just `.btn`:

```
.input-group {
    .btn {
        ...
    }
}
```
An [example from Bootstrap](https://getbootstrap.com/docs/5.1/forms/input-group/#button-addons):
```
<div class="input-group mb-3">
  <button class="btn btn-outline-secondary" type="button" id="button-addon1">Button</button>
  <input type="text" class="form-control" placeholder="" aria-label="Example text with button addon" aria-describedby="button-addon1">
</div>
```

azaslonov (changes requested):

azaslonov (changes requested):
Cristina, could you please include the same hide/reveal logic for Authorization header as well?

azaslonov (approved):

Issue #1220, opened 2021-03-31
Add content type request example selector to Test console.


azaslonov:
Currently, the Test console, when pre-filling the request body from representation example, takes the first request representation in the list, not allowing users to use another one, if it is present.

azaslonov:
Fixed in #1520.

PR #1520, opened 2021-11-09
#1220 Add content type request example selector to Test console

harunrst:
Added a selector to pick among representations to test console.

Here is the look;

Screen Shot 2021-11-09 at 16 00 23

**Note**: Added only for raw representations because only raw representations are allowed to add on Azure API Management.


src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -270,8 +270,19 @@ <h3>Headers</h3>
     <!-- ko if: hasBody -->
     <h3>Body</h3>
 
-    <div class="flex justify-content-end">
-        <!-- ko if: request.readonlyBodyFormat -->
+    <div class="row flex flex-row" style="margin-bottom: 3%;">
```
azaslonov, line 273: Please avoid using `style` attribute in the code and rely on the styles. I think what you may want to checkout is `.form-group` class. Another things is 3% - which seems unreliable taking into account customers may use customization tools to adjust padding/margins.

src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -270,8 +270,19 @@ <h3>Headers</h3>
...
+        <div class="col-4">
+            <!-- ko if: request.bodyFormat() === 'raw' -->
+            <select id="requestRepresentations" class="form-control" style="width: auto;" data-bind="value: $component.selectedRepresentation">
+                <!-- ko foreach: { data: request.representations, as: 'representation' } -->
+                <option data-bind="value: representation, text: 'Example ' + ($index() + 1)"></option>
```
azaslonov, line 278: `'Example ' + ($index() + 1)` - code samples come with a name which is reflected in operation details and hence the customers would expect to see same names in this dropdown.

azaslonov (changes requested):

mikebudzynski:
@harunrst, can you provide us with screenshots / explain:

1. What the initial state of the console is - we should preselect the first example from the dropdown
2. What happens if there are no examples - we should hide the dropdown
3. What happens if a user modifies the sample - we should probably change the option in the dropdown to something like "Custom" to allow users to revert back to the provided example


In addition to that, we should add a label to clarify what the dropdown does. For example, "Select a sample request body".

src/components/operations/operation-details/ko/runtime/operation-console.html
harunrst: Thank you I have followed your instructions and you can see it now :)

src/components/operations/operation-details/ko/runtime/operation-console.html
harunrst: it is the field called "typeName" which is not a required field so the user can still leave it empty. I updated the code as - using the name if exists, if not I am adding Example with its index. So that users can easily spot named representations and also not named ones.

harunrst:
> @harunrst, can you provide us with screenshots / explain:
> 
> 1. What the initial state of the console is - we should preselect the first example from the dropdown
> 2. What happens if there are no examples - we should hide the dropdown
> 3. What happens if a user modifies the sample - we should probably change the option in the dropdown to something like "Custom" to allow users to revert back to the provided example
> 
> In addition to that, we should add a label to clarify what the dropdown does. For example, "Select a sample request body".

mikebudzynski:
> here are the answers to your questions,
> 
> 1. Yes, it preselects the first example.
> 2. Yes, I hide the dropdown when there are no examples.
> 3. I am not sure if this would be clear for the user. Imagine you select an example and edit it, it created an example named Custom. Then user selected another example and edited it too, this created another example named Custom 1. And the user reselected the first example and edit it again and created another named Custom 2 and there are 3 examples of the same example now, and goes on. This could be hard to follow and **If the user needs to revert back to the initial state of the example, reselecting the example reverts it back, so the question here is "does the user need to save an updated example?".** Please tell me the way It should be so that I can code it right. Different rules might show up in this case.
> 4. I added the label

@harunrst, can we do something like this instead?

example selection dev portal

The "Revert" icon will show up only when the sample has been edited compared to the original sample from the API reference. We can display a tooltip when it's hovered with the message "The request body has been modified. Revert to the original request body sample.".

harunrst:
@mikebudzynski thank you for the review, I am adding the revert icon with its action.
@azaslonov I couldn't find an easy way to add an icon(I will be using an icon that is not used before) button with a tooltip, I think I am missing something in the project, what is the easiest way to do add it? Because it is just a simple icon button and shouldn't be trying I think.

src/components/operations/operation-details/ko/runtime/operation-console.html
azaslonov: As far as I understand, there can be just one empty, which you may name just "Example" or "Default".

azaslonov: Also, if you use rely on ` $index()` here, you may have situation like this: ["MyExample", "AnotherExample", "Example 3"] - the numbering will be off.

mikebudzynski:
@harunrst, great - please, provide us with a screenshot when you add the icon and rearrange the items.

src/components/operations/operation-details/ko/runtime/operation-console.html
harunrst: No, It can be multiple unnamed representations. So that is why that made sense to me with index. ["Example 1", "my named example", "Example 3"]. What do you suggest here?

azaslonov: Yep, saw the case you sent me offline, makes sense then.

src/models/console/consoleRequest.ts
```diff
@@ -12,13 +12,15 @@ export class ConsoleRequest {
...
+    public readonly selectedRepresentation: ko.Observable<ConsoleRepresentation>;
     public readonly description: string;
     public readonly body: ko.Observable<string>;
+    public readonly isBodyEdited: ko.Observable<boolean>
     public readonly hasBody: boolean;
     public readonly binary: ko.Observable<File>;
     public readonly bodyFormat: ko.Observable<RequestBodyType>;
     public readonly bodyDataItems: ko.ObservableArray<FormDataItem>;
-    public readonly representationContentType: string;
+    public readonly representationContentType: ko.Observable<string>;
```
azaslonov, line 23: `public readonly representationContentType: ko.Observable<string>` - this one is not used in the template, and you don't subscribe for its changes, is there a reason making it `ko.Observable`?

azaslonov (approved):
Added minor comment. Otherwise looks good.

src/models/console/consoleRequest.ts
harunrst: yes, but at line 44 I update it in a ko subscription, and if don`t define it as observable it doesn`t compile and says "it is a read-only field you can`t change it"

harunrst:
@mikebudzynski here is the last version, revert button was added with its functionality. 

image

mikebudzynski:
Thanks. Would be nice if we could fix the alignment to the middle inside the row. Currently the label, dropdown, and the icon seem to all be aligned differently.

harunrst:
@mikebudzynski I fixed the alignment. @azaslonov you can make the last review I guess If there is nothing else about UI/UX.

image

src/models/console/consoleRequest.ts
azaslonov: just remove `readonly` then :)

harunrst: ah just catch it :)

Issue #1041, opened 2020-11-18
HTML code injection


sbenji77:
## Bug description

Hi, When you try to insert an html tag, for example with toggle code, the tags are displayed in front after publishing it.

## Reproduction steps

1. Go to 'Developper portal'
2. Use text editor and select your tag
3. Click on 'Toggle code'
4. Publish and we can see the code in clear text in front

## Expected behavior

We would like to be able to insert small HTML snippets for example.

## Is your portal managed or self-hosted?

Self-hosted

On admin page
![image](https://user-images.githubusercontent.com/14352856/99588305-a6a74c00-29ea-11eb-9582-cc54542ffb77.png)

On front page 
![image](https://user-images.githubusercontent.com/14352856/99589635-91cbb800-29ec-11eb-9e7f-8732f24a5e07.png)

Thank you for your Help
Cordialy

azaslonov:
Hi @sbenji77, I actually just tried it, looks like it renders fine: 
![image](https://user-images.githubusercontent.com/2320302/99602798-f1a76c00-29b6-11eb-8597-03a97e5a2997.png)

Unless you want it to be rendered as:

- test1
- test2

If that's what you mean, then we don't support HTML injection into the page.

sbenji77:
It's a pity that it doesn't work because to customize some pages it would be a great simplification. Even the H tags at home don't work. Isn't this code toggle tool made for adding code? 
![image](https://user-images.githubusercontent.com/14352856/99651362-c8d6b380-2a56-11eb-873a-071fd78f925c.png)
On front
![image](https://user-images.githubusercontent.com/14352856/99651691-3be02a00-2a57-11eb-8f1b-38b2f80674e1.png)

Benjamin

mikebudzynski:
We have plans to create a widget that will allow HTML code injection. I will mark this Issue as a feature request to track the progress.

sbenji77:
Hello,

I would like to be able to add a modal, a slider, an anchor. I'm also looking for the possibility to put a search field in the pages as well but it's mainly for customisation.

Cordialy

AnRei123:
Hi, 

I would be also interested for the widget with the html code injection for the following use cases:
- create table and table styles
- allow to display of user-specific content based on the attached user groups. BUT the users should only be able to see information on APIs according to the assigned access groups in the APIM management.
[...]

PR #1526, opened 2021-11-11
HTML injection widget

JMach1:


mikebudzynski:
Referencing #1041

src/components/html-injection/ko/widgetEditorViewModel.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.onChange(this.model);
+    }
+
+    public async initMonaco() {
+        loader.config({ paths: { vs: "/assets/monaco-editor/vs" } });
+        loader.init().then(() => {
+            const settings: Record<string, unknown> = HtmlEditorSettings.config
+            settings.value = this.htmlCode() || '';
+
+            this[HtmlEditorSettings.id] = (window as any).monaco.editor.create(document.getElementById(HtmlEditorSettings.id), settings);
```
azaslonov, line 51: `document.getElementById(HtmlEditorSettings.id)` - technically there might be cases when 2 editors present in the view. In this case they will have same `id` and hence introduce logic and accessibility bugs. Using custom binding approach mentioned above should help with this issue, plus it will avoid breaking [MVVM model](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel), which Knockout (and other UI frameworks) use.

src/components/html-injection/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const widgetName = "html-injection";
+export const widgetDisplayName = "HTML injection";
```
azaslonov, line 2: I would change name "HTML injection" which doesn't look friendly for end-users. Maybe something like "Custom HTML block"?

azaslonov (changes requested):

src/components/html-injection/ko/widgetEditorViewModel.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private applyChanges(key: string): void {
+        this.model[key] = this[key]();
+        this.onChange(this.model);
+    }
+
+    public async initMonaco() {
+        loader.config({ paths: { vs: "/assets/monaco-editor/vs" } });
+        loader.init().then(() => {
```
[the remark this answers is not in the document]
azaslonov: Also, since you use `finally`, I guess you want to wrap it into `try`..`catch`..`finally` block.

src/components/html-injection/ko/widgetViewModelBinder.ts
```diff
@@ -0,0 +1,52 @@
...
+                name: widgetName,
+                displayName: widgetDisplayName,
+                readonly: bindingContext ? bindingContext.readonly : false,
+                model: model,
+                flow: "block",
+                editor: widgetEditorSelector,
+                draggable: true,
+                applyChanges: async () => {
+                    await this.updateViewModel(model, viewModel);
+                    this.eventManager.dispatchEvent("onContentUpdate");
```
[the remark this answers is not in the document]
JMach1: I've rebased this locally to latest origin/master, but I could not find `Events` nor `ContentUpdate` anywhere. Maybe it is in different branch?

src/components/html-injection/ko/widgetEditorView.html
```diff
@@ -0,0 +1,29 @@
+<fieldset class="form flex-item flex-item-grow">
+    <div class="form-group" style="display: flex; flex-direction: column; height: 100%;">
```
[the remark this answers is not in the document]
JMach1: I wanted the editor to be same height as the Paperbits window (and automatically resize together with the window).
Those styles are needed as without them it looks like this (but I'll rewrite it to use only classes):
image

src/components/html-injection/ko/widgetEditorView.html
```diff
@@ -0,0 +1,29 @@
...
+                data-bind="tooltip: 'Height in px of the <code>iframe</code> element the HTML code will be injected to.'"></button>
+        </label>
+        <div class="input-group">
+            <input id="height" type="number" class="form-control" placeholder="(Default of the browser)"
+                data-bind="textInput: htmlCodeHeight" />
+        </div>
+        <label for="htmlCode" class="form-label">
+            HTML code
+        </label>
+        <div class="input-group" style="height: 100%;">
```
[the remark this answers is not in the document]
JMach1: refactored

src/components/html-injection/widgetModelBinder.ts
```diff
@@ -0,0 +1,37 @@
...
+import { WidgetModel } from "./widgetModel";
+import { Contract } from "@paperbits/common";
+import { WidgetContract } from "./widgetContract";
+import { htmlCodeInitial } from "./ko/constants";
+
+/**
+ * This is a class that helps to prepare the model using data described
+ * in the contract.
+ */
+export class WidgetModelBinder implements IModelBinder<WidgetModel> {
```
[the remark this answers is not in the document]
JMach1: refactored

src/components/html-injection/ko/runtime/widget-runtime.html
```diff
@@ -0,0 +1,8 @@
+<iframe
+    data-bind="attr: { 
+        srcdoc: htmlCode,
+        style: `height: ${htmlCodeSizeStyles.height}; width: ${htmlCodeSizeStyles.width}`,
```
JMach1, line 4: @azaslonov you've mentioned some better way how this could be done

src/components/html-injection/ko/widgetEditorView.html
```diff
@@ -0,0 +1,11 @@
+<fieldset class="form flex flex-item flex-item-grow no-overflow">
+    <div class="form-group flex flex-column" style="display: flex; margin-bottom: 0;">
```
JMach1, line 2: @azaslonov unfortunately I was unable to find a way how to get rid of this inline style (using only existing classes), do you have any idea or I should just create a new class?

src/components/html-injection/ko/widgetEditorViewModel.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.htmlCodeHeight.subscribe(() => this.applyChanges('htmlCodeHeight'));
+        this.initMonaco();
+    }
+
+    private applyChanges(key: string): void {
+        this.model[key] = this[key]();
+        this.onChange(this.model);
+    }
+
+    public async initMonaco() {
```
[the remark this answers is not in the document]
JMach1: I've tested two editors in the same widget and it worked fine.

src/components/html-injection/ko/runtime/widget-runtime.html
azaslonov: Yep, there is a `styled` binding handler. You can see its usage example here: https://github.com/paperbits/paperbits-core/blob/master/src/button/ko/button.html, and filling the styles for it here: https://github.com/paperbits/paperbits-core/blob/master/src/button/ko/buttonViewModelBinder.ts#L36

src/components/html-injection/ko/widgetEditorView.html
azaslonov: Try to take a look at [this editor](https://github.com/paperbits/paperbits-styles/blob/master/src/workshops/styleEditor/styleEditor.html), it has all kinds of cases. If your case still doesn't fit into any of it, then yes, just create a new class.

src/components/html-injection/ko/widgetViewModelBinder.ts
```diff
@@ -0,0 +1,88 @@
...
+                htmlStyling += `<style>${compiler.compile(globalStyleSheet)}</style>`;
+
+                // Reflect style changes in designer
+                // TODO: looks like this is not needed after all
+                /* this.eventManager.addEventListener("onStyleChange", async (styleSheet) => {
+                    console.log({styleSheet});
+                    // viewModel.htmlStyling(styleSheet);
+                    // await this.updateViewModel(model, viewModel);
+                    // this.eventManager.dispatchEvent(Events.ContentUpdate);
+                }); */
```
JMach1, line 46: @azaslonov it appears it works just fine even without this event listener, changes from the Styles page are propagated correctly - I guess that was this event listener supposed to handle

src/components/html-injection/ko/htmlInjectionViewModelBinder.ts
```diff
@@ -0,0 +1,88 @@
...
+import { widgetName, widgetDisplayName, widgetEditorSelector } from "../constants";
+import { HtmlInjectionViewModel } from "./htmlInjectionViewModel";
+import { ViewModelBinder } from "@paperbits/common/widgets";
+import { StyleCompiler } from "@paperbits/common/styles";
+import { JssCompiler } from "@paperbits/styles/jssCompiler";
+import { ISettingsProvider } from "@paperbits/common/configuration";
+import { HTMLInjectionModel } from "../htmlInjectionModel";
+
+/**
+ * This class describes how the model needs to be presented (as a view model)
```
azaslonov, line 13: These comments left from tutorial:  `This class describes how the model needs to be presented (as a view model)...`

src/components/html-injection/ko/htmlInjectionViewModelBinder.ts
```diff
@@ -0,0 +1,88 @@
...
+
+            if (environment === "development") {
+                // Access global style sheet
+                const globalStyleSheet = await this.styleCompiler.getStyleSheet();
+
+                // Compile to CSS
+                const compiler = new JssCompiler();
+                htmlStyling += `<style>${compiler.compile(globalStyleSheet)}</style>`;
+
+                // Reflect style changes in designer
```
azaslonov, line 39: Can you please remove these comments? `// TODO: looks like this is not needed after all...`

src/components/html-injection/htmlInjectionContract.ts
```diff
@@ -0,0 +1,13 @@
+import { LocalStyles } from "@paperbits/common/styles";
+import { Contract } from "@paperbits/common";
+import { SizeStylePluginConfig } from "@paperbits/styles/plugins";
+
+/**
+ * Data contact (data layer) that defines how widget configuration gets persisted.
+ */
+export interface HTMLInjectionContract extends Contract {
+    htmlCode: string;
+    htmlCodeSizeStyles: SizeStylePluginConfig;
```
azaslonov, line 10: `htmlCodeSizeStyles: SizeStylePluginConfig;` - It's not in use anymore.

src/components/html-injection/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const widgetName = "html-injection";
```
azaslonov, line 1: `export const widgetName = "html-injection";` I would also rename it to `custom-html-code` to match display name.

src/components/html-injection/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+export const widgetDisplayName = "Custom HTML code";
+export const widgetCategory = "Advanced";
+export const widgetSelector = "widget";
```
azaslonov, line 4: `export const widgetSelector = "widget";` - this selector (and few others here) is generic (taken from tutorial) and may conflict with other widgets, if not renamed.

azaslonov (changes requested):

src/components/html-injection/htmlInjectionHandlers.ts
```diff
@@ -20,14 +13,11 @@ export class HTMLInjectionHandlers implements IWidgetHandler {
...
 
-            /**
-             * This method invoked when the widget gets added to the content.
-             */
             createModel: async () => { 
                 const model = new HTMLInjectionModel();
                 model.htmlCode = htmlCodeInitial;
-                model.htmlCodeSizeStyles = htmlCodeSizeStylesInitial;
                 model.inheritStyling = true;
+                StyleHelper.setPluginConfigForLocalStyles(model.styles, "size", htmlCodeSizeStylesInitial);
```
JMach1, line 20: @azaslonov I've noticed you did not init the model.styles with any initial value. I guess we should use the height 500px & width 100%?

src/components/html-injection/ko/htmlInjectionViewModelBinder.ts
```diff
@@ -49,7 +49,7 @@ export class HtmlInjectionViewModelBinder implements ViewModelBinder<HTMLInjecti
                 displayName: widgetDisplayName,
                 readonly: bindingContext ? bindingContext.readonly : false,
                 model: model,
-                flow: ComponentFlow.Contents,
+                flow: ComponentFlow.Block,
```
JMach1, line 52: @azaslonov I've noticed the "ComponentFlow.Contents" breaks the styling a little - 100 % width no longer makes the iframe 100 % of the parent. I thing the "Block" one is more suitable for this case, or did you change it on purpose?

src/components/html-injection/htmlInjectionHandlers.ts
azaslonov: Yes, you're right.

src/components/html-injection/ko/htmlInjectionViewModelBinder.ts
azaslonov: I guess I did it initially because we had a runtime widget and, hence, another wrapper element. As long as we can set both width and height, we should be good.

azaslonov: (feel free to resolve this one)

src/components/html-injection/htmlInjectionHandlers.ts
azaslonov: (feel free to resolve this one)

azaslonov (approved):

PR #1530, opened 2021-11-16
Fixed $ref displaying inside an array definition

jsorohova:


src/models/typeDefinition.ts
```diff
@@ -193,7 +193,14 @@ export class TypeDefinitionObjectProperty extends TypeDefinitionProperty {
         }
 
         if (contract.type === "array") {
```
azaslonov, line 195: Regarding this block:
```
        if (contract.type === "array") {
            if (contract.items.$ref) {
```

I think we should add check for presence `contract.items` as well, to minimize chance of breaking the page.

azaslonov (approved):
Just one comment. Otherwise looks good.

src/models/typeDefinition.ts
jsorohova: Will this check work? Because items are required for type array (accordingly to OpenAPI schema definition documentation).

PR #1535, opened 2021-11-22
Fixed: When adding a new page, the UI redirects to the newly created page

malincrist:
Fixed by this: https://github.com/paperbits/paperbits-common/issues/9

The issue was that one query was used for the searchRemote and also for the searchLocalState. SearchRemote was copying the query and then modifying the filter to fit what is on the server, but this didn't match with what is on the local.


azaslonov (approved):

jsorohova:
Next time commit package-lock file as well, please.

PR #1541, opened 2021-11-24
Fixed issues with example/examples/sample fields conflicts and rendering.

azaslonov:
- Removed duplicate examples rendering.
- Renamed "code-sample" component to "code-snippet".
- Fixed code-snippet title alignment issue.


src/apim.runtime.module.ts
```diff
@@ -94,7 +94,7 @@ export class ApimRuntimeModule implements IInjectorModule {
         injector.bind("graphqlConsole", GraphqlConsole);
         injector.bind("authorization", Authorization);
         injector.bind("typeDefinition", TypeDefinitionViewModel);
-        injector.bind("codeSample", CodeSampleViewModel);
+        injector.bind("codeSnippet", CodeSnippet);
```
malincrist, line 97: Shouldn't we keep the naming convention and name this class `CodeSnippetViewModel` ? 
Or should we agree that we drop the `ViewModel` naming and make this uniform among the application?

src/components/operations/operation-details/ko/runtime/type-definition.ts
```diff
@@ -62,11 +62,6 @@ export class TypeDefinitionViewModel {
         this.name(this.definition.name);
         this.description(this.definition.description);
         this.kind(this.definition.kind);
-
-        if (this.definition.example) {
-            this.exampleFormat(this.definition.exampleFormat);
-            this.example(this.definition.example);
-        }
```
malincrist, line 69: I don't think `exampleFormat` and `example` are used anywhere else. Maybe we could also remove their declarations?

src/apim.runtime.module.ts
azaslonov: There is no actual naming convention in place, but that's a good point to have one. I have added a work item for us to address it later.

src/components/operations/operation-details/ko/runtime/type-definition.ts
azaslonov: Indeed. Cleaned up.

ygrik (approved):

PR #1553, opened 2021-12-02
Remove item icon and OAuth sign-in and sign-up message

jsorohova:


src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -140,8 +140,8 @@ <h3>Parameters</h3>
                     </div>
                 </div>
             </div>
-            <div class="col-2">
-                <a href="#" role="button" data-bind="click: $component.removeQueryParameter">Remove</a>
+            <div class="col-1">
+                <a href="#" role="button" data-bind="click: $component.removeQueryParameter" class="remove-button remove-inline" aria-label="Remove parameter"><i class="icon-emb icon-emb-trash"></i></a>
```
azaslonov, line 144: `class="remove-button remove-inline"` - I have a suspicion that this could be `no-border` class instead of these two new classes. Please take a look at [authorizaion.html](https://github.com/Azure/api-management-developer-portal/blob/master/src/components/operations/operation-details/ko/runtime/authorization.html#L3) and how this component looks like in the console.

There is a small with your current approach: This way we introduce pattern of adding a new CSS class for each new action (in this case "remove-button"). As mentioned above, we may need just a style variation like "no-border" or "btn-no-border" (or something like this) that would apply to all similarly looking buttons (Please take a look at the Bootstrap approach for buttons [here](https://getbootstrap.com/docs/5.1/components/buttons/)).

src/themes/website/styles/forms.scss
```diff
@@ -281,4 +281,14 @@ $search-button-width: 30px;
 }
 .input-password::-ms-reveal {
     display: none;
-}
\ No newline at end of file
+}
+
+.remove-button {
```
azaslonov, line 286: Since this style related to buttons, I'd suggest to move it to `buttons.scss` (but please read the previous comment first).

src/themes/website/styles/widgets/textblock.scss
```diff
@@ -25,3 +25,10 @@ p:empty {
         position: relative;
     }
 }
+
+.description-text {
```
azaslonov, line 29: A few comments here:
- There is `.not-configured` class already (it's used i.e. in Menu widget for similar purpose).
- `textblock.scss` contains "Text block" widget specific styles, this one looks more like for `utils.scss` file.
- if you assign a `background-color`, you may need to assign `color` as well, otherwise the block will inherit it from global stylesheet where customers can assign text colors that will be poorly visible on this background.

src/themes/website/styles/forms.scss
```diff
@@ -281,4 +281,14 @@ $search-button-width: 30px;
...
+    display: flex;
+    align-items: center;
+    height: 100%;
+}
+
+.remove-inline {
+    padding-bottom: 1rem;
```
azaslonov, line 293: `padding-bottom: 1rem;` - are you sure about using `rem` here? This basically means that your padding will use the root element font size (some [useful reading](https://medium.com/@madhum86/css-font-sizing-pixels-vs-em-vs-rem-vs-percent-vs-viewport-units-b1485716afe7)).

src/components/users/signup-social/ko/signupSocial.html
```diff
@@ -1 +1,12 @@
+<!-- ko if: identityProvider -->
 <signup-social-runtime></signup-social-runtime>
+<!-- /ko -->
+
+<!-- ko ifnot: identityProvider -->
+<div class="description-text">This widget will display a sign-up form when you configure <a
+    href="https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-aad" target="_blank"
```
azaslonov, line 7: You should be able to add links to https://aka.ms - it's better to do that because the documentation moves from time to time and if we have a short link, we won't need to update application, only the short link itself in aka.ms tool. At least it worth moving it from the markup to a constant, and reference it here.

I have added two short links:
https://aka.ms/apim-how-to-aad
https://aka.ms/apim-how-to-aadb2c

azaslonov: `alt` attribute is useful only in  tag.

azaslonov (changes requested):

azaslonov (approved):

PR #1559, opened 2021-12-06
Fixed issue with redirect after a successful sign-in

JMach1:


src/startup.runtime.ts
```diff
@@ -16,7 +16,8 @@ document.addEventListener("DOMContentLoaded", () => {
 
 window.onbeforeunload = () => {
     if (!location.pathname.startsWith("/signin-sso") &&
-        !location.pathname.startsWith("/signup")) {
+        !location.pathname.startsWith("/signup") &&
+        !location.pathname.startsWith("/signin")) {
         const rest = location.href.split(location.pathname)[1];
         const returnUrl = location.pathname + rest;
         sessionStorage.setItem("returnUrl", returnUrl);
```
JMach1, line 23: The issue is that on page reload it saves "/signin" to "returnUrl" and after sign-in it redirects to it. I was checking the usage of "returnUrl" and I think, this should be fine but I'm not sure if it does not break something somewhere... I fix it by moving this check to the sign-in component, but I think this current solution is better.

ygrik (approved):

PR #1584, opened 2022-01-05
Create pipeline that will automatically check for Accessibility Issues

malincrist:
~~Work in progress, don't merge~~

Added pipeline that will automatically check for accessibility issues (using the Microsoft Accessibility Insights Github Action: https://github.com/microsoft/accessibility-insights-action)
Added static data(defaultStaticData.json), so the application can run and the scan can be performed without connecting to any backend service.

Fixed the accessibility issues found by the scan.


tests/mocks/defaultMockData.json
```diff
@@ -0,0 +1,4701 @@
...
+        "statusCode": 200,
+        "statusText": "OK",
+        "body": {
+            "aad": {
+                "clientId": "e87c98aa-6f4b-4a72-95d0-fc830a64ee06",
+                "authority": "login.microsoftonline.com"
+            }
+        }
+    },
+    "/auth-servers.json": {
```
azaslonov: Since it's just accessibility testing, we could reduce this data:
1. We can leave here just 1 fake OAuth server, and replace all there URLs and client IDs with some random data like `000-000-000`.
2. Same for APIs, just a couple of them should be enough, i.e. Echo API and a GraphQL API (@javierbo1989 can you please give Cristina a very simple one with just a few types in it, not SpaceX :)).
3. All these headers below are not necessary, we can leave just empty array.

azaslonov (changes requested):
Cristina, could you please revert default.json file? You've got another one for testing anyways.

tests/mocks/defaultMockData.json
malincrist: Good points, I will make these changes to reduce the data :)
Regarding the headers, some responses require the "content-type" header to know how to de-serialize the response, but we can remove the rest of the headers indeed.

src/components/users/profile/ko/runtime/profile.html
```diff
@@ -1,4 +1,6 @@
 <div class="panel panel-table">
+    <img
```
harunrst, line 2: what is it for?

src/services/IStaticDataProvider.ts
```diff
@@ -0,0 +1,3 @@
+export interface IStaticDataProvider {
```
harunrst, line 1: not important just to say, if you only export this, you can export like 'export default interface IStatiicDataProvider'. Then you can import it like 'import IStatiicDataProvider from "./services/staticDataHttpClient". instead import { StaticDataHttpClient } from "./services/staticDataHttpClient";

src/startup.publish.ts
```diff
@@ -10,11 +10,23 @@ import { StaticSettingsProvider } from "./components/staticSettingsProvider";
...
+if (process.env.NODE_ENV === "staticData") {
+    settingsProvider = new StaticSettingsProvider({
+        "environment": "publishing",
+        "managementApiUrl": "https://contoso.management.azure-api.net",
+        "managementApiAccessToken": "SharedAccessSignature&1&",
+        "useHipCaptcha": false
+    });
+}
+else {
+    const configFile = path.resolve(__dirname, "./config.json");
```
harunrst, line 26: don't you need to set config files in staticData mode as well?

webpack.staticData.js
```diff
@@ -0,0 +1,19 @@
+const webpack = require("webpack");
+const publisherConfig = require("./webpack.publisher");
+
+
+let staticData = publisherConfig;
+
+staticData.publisherConfig.mode = "none";
+staticData.publisherConfig.plugins.push(new webpack.DefinePlugin({
+    "process.env.NODE_ENV": JSON.stringify("staticData")
```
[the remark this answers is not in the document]
harunrst: it could be better to have const definitions for environments because you use this static string in another file as well. so that it could be a single source, easy to read, and fault-tolerant.

src/components/users/profile/ko/runtime/profile.html
malincrist: I just added something un-accessible to see if the scan will catch it. I have to delete it :D

webpack.staticData.js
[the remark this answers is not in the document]
malincrist: It needs to be escaped as that's how webpack checks it:
image

src/startup.publish.ts
malincrist: It's set inside the if, but it's not a config file, but the configuration directly as an object:
 settingsProvider = new StaticSettingsProvider(**{
        "environment": "publishing",
        "managementApiUrl": "https://contoso.management.azure-api.net",
        "managementApiAccessToken": "SharedAccessSignature&1&",
        "useHipCaptcha": false
    }**);

tests/mocks/defaultMockData.json
malincrist: Also, regarding the default.json file I have made changes there - there were some accessibility issues with color contrast that were fixed inside the default.json. But I have made it compact back.

environmentConstants.js
```diff
@@ -0,0 +1,6 @@
+
+const staticDataEnvironment = "staticData";
```
harunrst, line 2: mock environment could be the best naming for this purpose I guess, and we can also update the command as " npm run build-mock". What do you think?

src/services/publisherStaticDataProvider.ts
```diff
@@ -0,0 +1,24 @@
+import path = require("path");
+import  IStaticDataProvider  from "./IStaticDataProvider";
+const fs = require('fs');
+
+export class PublisherStaticDataProvider implements IStaticDataProvider {
+    getStaticData(objectType: string): Promise<any> {
+        let fullPath = objectType == "defaultMockData.json" ?
+            path.join(__dirname, '../../tests/mocks/defaultMockData.json') : path.join(__dirname, '../../templates/default.json');
+        let filePath = path.resolve(fullPath);
+        console.log(`fullPath: ${fullPath} and filepath: ${filePath}`);
```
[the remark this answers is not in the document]
malincrist: forgotten, I removed it

environmentConstants.js
malincrist: I'm not sure: "Mock. Mocks are the objects that store method calls. It referred to as the dynamic wrappers for dependencies used in the tests."
I think a better idea would be to change the defaultMockData.json file to defaultStaticData.json, so we keep the same naming.

harunrst: hmm, I am fine with that as well, it is your decision then :)

harunrst (approved):

Issue #935, opened 2020-09-21
OpenAPI schema issue when using polymorphism 


konsta2106:
## Bug description

Hello. I'm using polymorphism in OpenAPI spec. In case of Swagger when it generates documentation, i can actually explore and expand all schemas used by polymorphism. In case of Azure APIM and developer portal, i can only see references to models, but i can't click and expand and explore schemas.

## Reproduction steps

1. Go to developer portal 'https://xxxxxxxxx.developer.azure-api.net/'
2. Click on 'Explore APIs' and select API. Select operation that is using polymorphism.
3. Scroll down to needed method that is using polymorphism and choose "schema view"
4. See error -> We can see references to models/schemas used by oneOf/anyOf but we cant see any of schemas for those references (links are not clickable)
![Screenshot 2020-09-17 at 10 19 06](https://user-images.githubusercontent.com/54150208/93765976-6ad84c00-fc1e-11ea-8a9a-d932d24f4d48.png)

## Expected behavior

I would expect same kind of behaviour that for example Swagger does.
![Screenshot 2020-09-17 at 10 13 09](https://user-images.githubusercontent.com/54150208/93766744-8bed6c80-fc1f-11ea-9c7d-12abf3ce8043.png)
And after i click on one of them it will expand schema:
![Screenshot 2020-09-21 at 15 26 42 copy](https://user-images.githubusercontent.com/54150208/93766761-93147a80-fc1f-11ea-8dd6-66d46a4164bb.png)

## Is your portal managed or self-hosted?

Managed

## Release tag or commit SHA (if using self-hosted version)

none

## API Management service name

Contoso

## Environment

 - Operating system: [mac OS]
 - Browser: [Google Chrome]

azaslonov:
Hi Konstantin, we'll discuss this and see what we can do. Thank you.

konsta2106:
Hello @azaslonov. Is there any update on this?

konsta2106:
Hi @azaslonov. It Would be nice if there is any information or update about this feature since it is pretty crucial for our partners to see what properties are needed by different schemas and to see also descriptions to every property. At this point there is no any information with straight reference to oneOf or anyOf schema and they can not be discovered anywhere in portal. Only one pre-defined example from one schema is visible

![Screenshot 2020-12-01 at 14 54 17 copy](https://user-images.githubusercontent.com/54150208/100744114-f839d880-33e5-11eb-9f83-5b2540babbac.png)
---
[...]

Issue #1248, opened 2021-04-13
Implement rendering of top-level "allOf" definition as a combination of properties of enumerated definitions


azaslonov:
We should be able to properly render "allOf" combination in API documentation, so that schema like:
```
SomeObject:
   allOf:
      $ref: '#/components/schemas/SomeChildObject'
      type: object
      properties:
         propertyA:
           type: string
         propertyB:
            type: string

SomeChildObject 
   type: object
   properties:
      propertyX:
         type: string
```

Would be rendered as if we had object like:
```
SomeObject 
   type: object
   properties:
      propertyA:
         type: string
      propertyB:
         type: string
      propertyX:
         type: string
```

![image](https://user-images.githubusercontent.com/2320302/114634037-8f594b80-9c76-11eb-9144-69e4d6d0edf4.png)

cosminstirbu:
@jsorohova - I've noticed you've closed this issue (and the related bug https://github.com/Azure/api-management-developer-portal/issues/1366) . However, it seems that it is still reproducing.

Have the changes actually been deployed?

cosminstirbu:
This is how it looks on our end (disregard Tenant-tenant-Analyze-syncPostRequest which is tracked via https://github.com/Azure/api-management-developer-portal/issues/1192)
![Screenshot 2022-03-15 at 15 25 42](https://user-images.githubusercontent.com/7135690/158387493-cf815c5a-6bd6-4bac-a561-89ce504c3a08.png)

cosminstirbu:
Is there a private way to do that?

jsorohova:
You can email me at jsorohova@microsoft.com

Issue #669, opened 2020-06-02
OneOf problems


abusse-interhyp:
## Bug description

I know there was a fix regarding oneOf in the last release. I still have some problems with it or at least some questions.

## Reproduction steps
See attached yaml files
[api.zip](https://github.com/Azure/api-management-developer-portal/files/4715431/api.zip)
in the oneOf.yml Cat and Dog are explicitly linked via an oneOf option.
In the baseClass.yml both are indirectly linked based on the super type Pet in the patch Request. 

## Expected behavior
The Objects Cat and Dog should be visible in the Developer portal at least for the oneOf.yml.
Based on other tools we are currently using, it would be perfect if they are also visible for the baseClass.yml.
There are some other tools on the market, some of them meet this criteria, some others not.
Is it possible for the API Management?

## Is your portal managed or self-hosted?

Managed (and currently republished)

## API Management service name

Region: West Europe

azaslonov:
Hi @abusse-interhyp, sorry for the delay, we're looking into that.

albusse:
@azaslonov thabk you 
Our actually planed API is based on the baseClass.yml with other data and more nested. 
It‘s still in private preview but if you need more examples please let me know.

abusse-interhyp:
Any updates on the timeline for this issue?
Our current APIs are using the polymorphism on the first levels, therefore many DTOs are not shown in the Developer Portal atm.

The baseClass.yml approach I mentioned above is now working with the other tools we are using (generators and internal swagger ui) so this would be the preferred behaviour.

Thank you.

abusse-interhyp:
@mikebudzynski @azaslonov 
Hi Guys, 
any update on this?

Actually it is a showstopper for us at the moment and brings up too many questions from our clients, because so many DTOs are not shown in the Api description page.

Just to clarify, will it be possible to have some functionality similar to the following screenshots?

![image](https://user-images.githubusercontent.com/64847387/105141308-3eafd780-5af9-11eb-90c0-0838ea4a488e.png)
![image](https://user-images.githubusercontent.com/64847387/105141364-52f3d480-5af9-11eb-9c66-e5aaa291a8b3.png)
![image](https://user-images.githubusercontent.com/64847387/105141438-6ef77600-5af9-11eb-9137-c3113407bade.png)

based on a yaml file like this:
https://raw.githubusercontent.com/Interhyp/Top-Rate-API/master/top-rate.yaml

Thank you for an answer.

[...]

PR #1601, opened 2022-01-19
Complex schemas documentation rendering

jsorohova:
Closes: #935 , #1248, #669

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/type-definition-combination.html
+++ b/src/components/operations/operation-details/ko/runtime/type-definition-combination.html
@@ -1,12 +1,14 @@
 <!-- ko if: properties && properties.length > 0 -->
 
-<h5 class="combination-type"><span data-bind="text: type.combinationType"></span>:</h5>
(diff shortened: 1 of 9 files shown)


src/components/operations/operation-details/ko/runtime/type-definition-combination.html
```diff
@@ -0,0 +1,37 @@
+<!-- ko if: properties && properties.length > 0 -->
+
+<h5 class="combination-type"><span data-bind="text: type.combinationType"></span>:</h5>
```
azaslonov, line 3: This style assigns margins, but at the same time users have control over H1-H6 sizes, so the will override this one. I'd rather wrapped each such section into DIV and assigned the margin to it.
Another thing, we're collecting that kind of utility classes in utils.scss file. Having it all in one place will help us identify patterns and adjust as we go.

azaslonov (changes requested):

azaslonov (approved):

PR #1623, opened 2022-01-27
Feature/graphql documentation

javierbo1989:


src/themes/website/styles/docs.scss
```diff
@@ -480,3 +484,43 @@ oauth-server-configuration {
         margin-top: 20px;
     }
 }
+
+.details-button {
```
azaslonov, line 488: Looking at usage of `.details-button` and `.clickable`, it seems that it's artifacts from previous implementation. Do we still need these classes?

Also, `clickable` is always a link, we should allow user to define the color. With this class you override whatever colors a user sets.

src/components/operations/operation-details/ko/runtime/operation-details.html
```diff
@@ -326,21 +326,19 @@ <h3>Protocol</h3>
 <!-- ko if: apiType() === 'graphql' -->
 <div class="animation-fade-in">
     <div class="operation-header">
-        <h2 class="operation-name"></h2>
+        <div>
+            <h2 class="gql-details">GraphQL Endpoint</h2>
```
azaslonov, line 330: The class `.gql-details` overrides the sizes defined by users. We should do it carefully or don't do it at all (also, we plan to add local styling to complex components, so that users will be able to adjust those as they like).

src/themes/website/styles/navs.scss
```diff
@@ -75,6 +75,10 @@
     .nav-item {
         flex: 100%;
     }
+
+    &.margin-top {
```
azaslonov, line 79: `.margin-top` could you please move this class to `utils.scss` and maybe rename it by analogy with similar classes. The problem is that we (all of us working on the UI) keep introducing same or similar classes and scatter them across the style files. Plus that meaning of those class names doesn't usually follow any pattern. Having them in one place at least gives us opportunity to find those patterns. Best of all, if we all learn about BEM model, which addresses exactly this problem. That would minimize usage of these utility classes.

azaslonov (approved):

Issue #674, opened 2020-06-02
Improvement: Add the Terms of Service to the Complete Sign Up Page (AAD)


VWACBrick:
This is similar to request to have the TOU added to the Basic Auth sign up page. 
#228 

**Azure Settings**
![image](https://user-images.githubusercontent.com/65634611/83577632-93aac900-a4e9-11ea-8504-b9a3beebcb76.png)

Currently our site is only using the AAD today but later will use B2C Authentication, can we have the same TOU feature surface on the Sign in Page and the "Complete Sign Up" page?

New Site Complete Sign Up page
![image](https://user-images.githubusercontent.com/65634611/83577601-783fbe00-a4e9-11ea-9393-f3a28986ed1f.png)

Legacy Portal during AAD only sign in
![image](https://user-images.githubusercontent.com/65634611/83577712-cbb20c00-a4e9-11ea-8d9b-472c6150560a.png)

New Site sign in page with AAD only (B2C coming soon)
![image](https://user-images.githubusercontent.com/65634611/83577761-e8e6da80-a4e9-11ea-886d-17723ffb2acb.png)

Thank you

azaslonov:
Hi @VWACBrick, makes sense, thanks for the suggestion.

khigashi1021:
Hi I have the customer who would like to have this improvement applied. Are there any plan to improve this?

mikebudzynski:
We do not have ETA at the moment, but I have prioritized the issue.

PR #1648, opened 2022-02-10
Added Terms of Use to all authentication pages

jsorohova:
Closes: #674


src/themes/website/styles/buttons.scss
```diff
@@ -7,3 +7,7 @@
 .btn-inline {
     padding-bottom: 1rem;
 }
+
+.btn-link {
+    text-decoration: underline;
```
azaslonov, line 12: `text-decoration: underline;` - Actually, we should not do this, because users should be able to redefine link appearance.

src/components/users/signin/ko/runtime/signin.html
```diff
@@ -14,6 +14,10 @@
                 data-bind="textInput: password">
         </div>
 
+        <!-- ko if: termsEnabled && termsOfUse -->
+        <terms-of-use params="{ isConsentRequired: isConsentRequired, consented: consented, termsOfUse: termsOfUse }"></terms-of-use>
```
azaslonov, line 18: I think the request was about "Complete signup" view, hence it should be only signupSocial.html. Currently, we can't enforce consent on sign-in, since it requires backend support for that, but for creating account we do support it now.

azaslonov (approved):

PR #1664, opened 2022-02-21
Added options to show/hide code samples

### LINKED ISSUE #1596: Code sample & large JWT reduces usability of test console
When using the test console with OAuth a HTTP code sample is automatically shown. As the token can be sizeable it can result in a significant amount of scrolling for a user to reach the send button and review the parameters to the request. Would it be possible to improve this situation? Could a code sample not be shown by default?

malincrist:
Closes #1596 

Made all sections in the test console collapsible.
Made the "send" button stick to the bottom of the console.
rest_api
ws_api


harunrst (approved):

azaslonov (changes requested):
Cristina, can we please re-use the collapsible section concept like here?
image

Plus that the code samples dropdown could go under the "HTTP request" header. This way we'll have it collapsed but keep the section heading visible.

Another thing I would ask (since we touch this functionality) is to make all the reset sections (Parameters, Headers, and Body) collapsible similarly.

azaslonov (approved):

PR #1688, opened 2022-03-10
Added support for union types in the graphql console

javierbo1989:
- Creating new nodes for the different types of the union.
- Using fragments ( ... on ) in the query editor for union types queries.
- Synchronized explorer and query editor for union types.
- Considered the highly used __typename Meta field in union types queries.


src/components/operations/operation-details/ko/runtime/graphql-console.ts
```diff
@@ -201,20 +194,20 @@ export class GraphqlConsole {
             // Go through every node in a new generated parsed graphQL, associate the node with the created tree from init and toggle checkmark.
             GraphQL.visit(ast, {
                 enter(node) {
-                    if (node.kind === "Field" || node.kind === "Argument" || node.kind === "ObjectField") {
+                    if (node.kind === "Field" || node.kind === "Argument" || node.kind === "ObjectField" || node.kind === "InlineFragment") {
```
azaslonov, line 197: Javier, could you please move those to constants. There is a chance that they already defined by the library, if not, let's just add it.

azaslonov: I mean all of those: `InlineFragment`, `Argument`, `Field`, `StringValue`, `Variable`, etc,

azaslonov (changes requested):

src/components/operations/operation-details/ko/runtime/graphql-console.ts
javierbo1989: Sure thing!

azaslonov (approved):

PR #1695, opened 2022-03-18
Add "API: Products (dropdown)" widget

### LINKED ISSUE #1177: "API: products (dropdown)" widget missing
## Bug description/proposed behavior

Currently there are only two API: product widgets:
- API: products
- API: products (tiles)

A "API: products (dropdown)" widget is missing.

The missing widget should perfectly compliment the view of the already existing widget "List of products".
image

The missing "API: products (dropdown)" widget should be directly placed together with the shown widget in the screenshot. The reference to the parent API products should be accessible from the context of the API description in the header section of the API details page.
For this use case tiles or tables are not suitable.

Remarks: as a temporary solution, we added the "API: products (table)" widget to the end of the API details page below the details of the operations. The table is hard to access, the user has to scroll a lot. A link on the same page to the bottom is currently not supported with the current hyperlink implementation. If we link to the table at the end of the page, the page is newly loaded and the table is empty. For details refer to ticket #1176 

It is urgent that either the link restriction or this ticket will be resolved as soon as possible to provide an acceptable developer experience.

From the API details page, only a list with references to associated parent APIs is required and a prominent search field and detailed descriptions of the API product. To maintain readability and to not overwhelm the reader with Information only the meaningful name should be displayed in the list.

Managed

mrcarlosdev:
## 🎩 Goal
Create a new widget to display in a dropdown format the products assigned to a specific API within the Developer Portal. Currently there are two widgets accomplishing this action (list and tiles) but the dropdown view was missing.

## 🗿 In the mid time
This feature was implemented by following the rules of the already existing widgets List of APIs (dropdown) and List of products (dropdown). So you can expect the same dropdown behavior (including the search within the dropdown option).

### Widget selector
drawing

### Widget selected

drawing

### Widget dropdown

drawing

### Widget filter

drawing

### Widget filter selected

drawing

Closes #1177

### DIFF
--- a/src/components/apis/api-products/ko/runtime/api-products-dropdown.ts
+++ b/src/components/apis/api-products/ko/runtime/api-products-dropdown.ts
@@ -1,0 +1,144 @@
+import * as ko from "knockout";
+import * as Constants from "../../../../../constants";
+import template from "./api-products-dropdown.html";
+import { Component, RuntimeComponent, OnMounted, Param, OnDestroyed } from "@paperbits/common/ko/decorators";
+import { Router } from "@paperbits/common/routing";
+import { ApiService } from "../../../../../services/apiService";
+import { SearchQuery } from "../../../../../contracts/searchQuery";
+import { RouteHelper } from "../../../../../routing/routeHelper";
+import { Product } from "../../../../../models/product";
+
+@RuntimeComponent({
+  selector: "api-products-dropdown-runtime",
+})
+@Component({
+  selector: "api-products-dropdown-runtime",
+  template: template,
+})
+export class ApiProductsDropdown {
+  public readonly products: ko.ObservableArray<Product>;
+  public readonly selectedApiName: ko.Observable<string>;
+  public readonly working: ko.Observable<boolean>;
+  public readonly pattern: ko.Observable<string>;
+  public readonly pageNumber: ko.Observable<number>;
+  public readonly totalPages: ko.Observable<number>;
+  public readonly selectedProduct: ko.Observable<Product>;
+  public readonly selectedProductName: ko.Observable<string>;
+  public readonly selection: ko.Computed<string>;
+
+  constructor(
+    private readonly apiService: ApiService,
+    private readonly router: Router,
+    private readonly routeHelper: RouteHelper
+  ) {
+    this.detailsPageUrl = ko.observable();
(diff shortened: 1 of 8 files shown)


src/components/apis/api-products/ko/runtime/api-products-dropdown.ts
```diff
@@ -0,0 +1,144 @@
...
+      const pageOfProducts = await this.apiService.getApiProductsPage(apiName, query);
+
+      this.products(pageOfProducts.value);
+      this.totalPages(Math.ceil(pageOfProducts.count / Constants.defaultPageSize));
+
+      const productName = this.routeHelper.getProductName();
+
+        this.selectedProduct(productName
+            ? pageOfProducts.value.find(item => item.id.endsWith(productName))
+            : pageOfProducts.value[-1]);
```
Mielek, line 126: What does it mean `pageOfProducts.value[-1]` ? if value is an array which I suppose it is, then it would return `undefined`. Why not just return write undefined?

On the other hand, If undefined do not need to be set to ko.Observable, then you can do it like this
```suggestion
            if(productName){
                this.selectedProduct(pageOfProducts.value.find(item => item.id.endsWith(productName)));
             }
```
Or in one line
```suggestion
            productName && this.selectedProduct(pageOfProducts.value.find(item => item.id.endsWith(productName)));
```

azaslonov (approved):

src/components/apis/api-products/ko/runtime/api-products-dropdown.ts
```diff
@@ -0,0 +1,144 @@
...
+  @Param()
+  public detailsPageUrl: ko.Observable<string>;
+
+  @OnMounted()
+  public async initialize(): Promise<void> {
+    const apiName = this.routeHelper.getApiName();
+
+    if (apiName) {
+      this.selectedApiName(apiName);
+      await this.loadPageOfProducts();
```
malincrist, line 62: Since `resetSearch()`  from the next line will call `loadPageOfProducts()` , is  this call needed here?

src/components/apis/api-products/ko/runtime/api-products-dropdown.ts
mrcarlosdev: Agreed. Thanks for the heads up.

PR #1698, opened 2022-03-22
custom-html widget iframe allow more attributes

JMach1:
### The solution
I've added following attributes:
```
allow-popups-to-escape-sandbox 
allow-top-navigation 
allow-presentation 
allow-orientation-lock 
allow-pointer-lock
allow-downloads 
```
I made it as a separate constant so it can be reused elsewhere, for example custom widget iframes.


azaslonov (approved):

rafaeleloi:
@azaslonov @JMach1 could you please inform if is this already available? I’ve seen that the PR seems to have been merged into the master branch, does that mean it is released as well? 

Do we need to do anything more than setting target=”_blank” on the links and then publish the portal?
Thanks.

PR #1703, opened 2022-03-28
Removed duplicated call to /apis for the API list widget

malincrist:
### **Problem:**
When loading the API list widget, there is a double call made to /apis.
In case the APIs are filtered after a tag, one of the calls will get and show the list with all the APIs and only the second one will get and show the filtered list of APIs:
image

The problem was that the components that used the component tag-input, during the initialization, were calling `loadPageOfApis()` twice: 
- once by a direct call from the `resetSearch()` method
- once by a subscription notification that the tags were changed

For the second call, the notification was only sent after the tag-input component finished loading and the tag list was populated. In case there was no filter applied to the API list, this notification was not send and so the `resetSearch()` call was needed.

### **Solution:**
Notify the components that use tag-filter that the tag list was loaded (also when it is empty), so they can make the call to retrieve the API list accordingly
Remove the direct call to `resetSearch()` from the initialization.

### DIFF
--- a/src/components/tag-input/tag-input.ts
+++ b/src/components/tag-input/tag-input.ts
@@ -65,6 +65,7 @@ export class TagInput {
         const tagsValue = this.routeHelper.getTags();
 
         if (!tagsValue) {
+            this.onChange([]);
             return;
         }
 
--- a/src/components/apis/list-of-apis/ko/runtime/api-list-dropdown.ts
+++ b/src/components/apis/list-of-apis/ko/runtime/api-list-dropdown.ts
@@ -58,7 +58,6 @@ export class ApiListDropdown {
 
     @OnMounted()
     public async initialize(): Promise<void> {
-        await this.resetSearch();
         await this.checkSelection();
 
         this.pattern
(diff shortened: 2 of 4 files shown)


src/components/tag-input/tag-input.ts
```diff
@@ -65,6 +65,7 @@ export class TagInput {
         const tagsValue = this.routeHelper.getTags();
 
         if (!tagsValue) {
+            this.onChange([]);
```
azaslonov, line 68: There should be a check if `onChange` is actually defined before invoking it. There is similar check below.

Also, probably this logic here need to be revised, because calling `onChange` when the widget just being initialized doesn't seem right.

azaslonov (changes requested):

src/components/tag-input/tag-input.ts
malincrist: Good point, I will make the check.

Regarding the logic:
I have tried to make a different event at first, `onLoaded()` that will notify the components only when the tag-input finished loading the list, but I noticed that it's exactly the same logic from the `onChange()` so I decided to use it, even if the name is not very suggestive for this case.

We can also make the api-list to get the tags and do the search according to them and send the tags to the tag-input, in order the component ( to be able to select tags from the UI)

malincrist: Good point, I will make the check.

Regarding the logic:
I have tried to make a different event at first, onLoaded() that will notify the components only when the tag-input finished loading the list, but I noticed that it's exactly the same logic from the onChange() so I decided to use it, even if the name is not very suggestive for this case.

We can also make the api-list to get the tags and do the search according to them and send the tags to the tag-input, in order the component ( to be able to select tags from the UI)

azaslonov (approved):

PR #1732, opened 2022-04-20
Cut off mapi requests

harunrst:
Problem
Most of the requests go directly to the MAPI Backend. This brings security vulnerabilities.

Solution
Getting rid of management API URLs and proxying requests through the dev portal backend.

IMPORTANT
This development should be considered with the dev portal backend's changes. Its PR is also open see the repository on Azure.


harunrst:
Changes in 

- accessTokenRefresher.ts
- mapiClient.ts
- usersService.ts

might affect self-hosted customers. See the usage of the function _getDeveloperEndpoint_ . This function append /developer at the end of the backendUrl when we want to proxy them. @mikebudzynski @azaslonov

PR #1743, opened 2022-04-27
Custom widget scaffolding tool

JMach1:


package.json
```diff
@@ -103,6 +104,7 @@
     "remark-parse": "^10.0.1",
     "remark-rehype": "^10.1.0",
     "saxen": "^8.1.2",
+    "scaffold": "scaffold-0.0.6.tgz",
```
JMach1, line 107: will be replaced just by version once the package is published in NPM

src/components/custom-widget/customWidgetContract.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface CustomWidgetContract extends Contract {
+    /**
+     * This name is used to identify which custom-widget-binder to load into iframe.
+     */
+    name: string;
+
+    /**
+     * @deprecated. This name will be shown in editors and custom-widget-binder selector. (this will be removed from here and used only in WidgetHandlers)
+     */
+    widgetDisplayName: string;
```
JMach1, line 13: @azaslonov you've marked the widgetDisplayName as deprecated and only to be used in WidetHandlers. As I'm using it even elsewhere, I would like to keep this property here if possible.

src/components/custom-widget-list/loadCustomWidgetConfigs.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (sources.length) {
+        sources.forEach(source => {
+            try {
+                const url = new URL(source);
+                overridesPromises.push(fetch(url.href + configFileName));
+            } catch (e) {
+                console.warn(source, e);
+            }
+        });
+    }
```
Mielek, line 38: Extract function

```ts
function fetchOverrides(sources: string[]) {
  const promises= [];
  if (sources.length) {
        sources.forEach(source => {
            try {
                const url = new URL(source);
                promises.push(fetch(url.href + configFileName));
            } catch (e) {
                console.warn(source, e);
            }
        });
    }
  return promises;
}
```

PR #1745, opened 2022-04-28
Local styles/product list

malincrist:
Enabled customizations for the following sub-components of the Product List widget:

Search component
- Text color
- Border color

 Grid 
- Background color 
- Lines color
- Text color

### DIFF
--- a/src/components/products/product-list/ko/runtime/product-list.html
+++ b/src/components/products/product-list/ko/runtime/product-list.html
@@ -25,7 +25,7 @@
                 <a data-bind="attr: { href: $component.getProductUrl(product) }, text: product.displayName"></a>
             </div>
             <div class="col-7">
-                <span data-bind="markdown: { source: product.description, truncateAt: 250 }"></span>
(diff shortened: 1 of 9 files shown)


src/components/products/product-list/ko/runtime/product-list.html
```diff
@@ -20,12 +20,12 @@
...
     <div class="table-body animation-fade-in">
         <!-- ko foreach: { data: products, as: 'product' } -->
-        <div class="table-row">
+        <div class="products-grid-row">
             <div class="col-5 text-truncate">
                 <a data-bind="attr: { href: $component.getProductUrl(product) }, text: product.displayName"></a>
             </div>
             <div class="col-7">
-                <span data-bind="markdown: { source: product.description, truncateAt: 250 }"></span>
+                <span class="cell-inherit-color" data-bind="markdown: { source: product.description, truncateAt: 250 }"></span>
```
malincrist, line 28: The markdown will surround the text with a `<p>` tag which has the color set globally from the styles; therefore, the typography color will be overwritten without this fix.

azaslonov (approved):

PR #1748, opened 2022-05-03
Wss protocol validations, building graphql path, added subscription-key to parameters.

javierbo1989:
- Ws(s) protocol validation for GraphQL ws subscriptions.
- Adding functionality to build the path for GraphQL operations via the http(s) and ws(s) protocols.
- Adding subscription-key parameter to path, for GraphQL APIs where subscription is required.

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/authorization.ts
+++ b/src/components/operations/operation-details/ko/runtime/authorization.ts
@@ -49,7 +49,8 @@ export class Authorization {
         this.authorizationServer = ko.observable();
         this.selectedGrantType = ko.observable();
         this.api = ko.observable<Api>();
-        this.headers = ko.observableArray<ConsoleHeader>();;
+        this.headers = ko.observableArray<ConsoleHeader>();
+        this.queryParameters = ko.observableArray<ConsoleParameter>();
         this.consoleOperation = ko.observable<ConsoleOperation>();
         this.templates = templates;
         this.codeSample = ko.observable<string>();
@@ -74,6 +75,9 @@ export class Authorization {
 
     @Param()
     public headers: ko.ObservableArray<ConsoleHeader>;
+
+    @Param()
+    public queryParameters: ko.ObservableArray<ConsoleParameter>;
 
     @Param()
     public codeSample: ko.Observable<string>;
@@ -380,8 +384,8 @@ export class Authorization {
             return;
         }
 
-        if (this.api().type === TypeOfApi.webSocket) {
-            this.setSubscriptionKeyParameter(subscriptionKey)
+        if (this.api().type === TypeOfApi.webSocket || this.isGraphQL()) {
+            this.setSubscriptionKeyParameter(subscriptionKey);
         } else {
             this.setSubscriptionKeyHeader(subscriptionKey);
         }
@@ -410,15 +414,20 @@ export class Authorization {
         keyParameter.required = true;
         keyParameter.inputType("password");
 
-        this.consoleOperation().request.queryParameters.push(keyParameter);
-        this.updateRequestSummary();
+        if (this.isGraphQL()) {
+            this.queryParameters.push(keyParameter);
+        }
+        else {
+            this.consoleOperation().request.queryParameters.push(keyParameter);
+            this.updateRequestSummary();
+        }
     }
 
     private getSubscriptionKeyParam(): ConsoleParameter {
         const subscriptionKeyParamName = this.getSubscriptionKeyParamName();
         const searchName = subscriptionKeyParamName.toLocaleLowerCase();
-
-        return this.consoleOperation().request.queryParameters().find(x => x.name()?.toLocaleLowerCase() === searchName);
+        let queryParameters = (this.isGraphQL()) ? this.queryParameters() : this.consoleOperation().request.queryParameters();
+        return queryParameters.find(x => x.name()?.toLocaleLowerCase() === searchName);
     }
 
     private getSubscriptionKeyParamName(): string {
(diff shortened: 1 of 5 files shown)


src/components/operations/operation-details/ko/runtime/authorization.ts
```diff
@@ -380,8 +384,8 @@ export class Authorization {
             return;
         }
 
-        if (this.api().type === TypeOfApi.webSocket) {
-            this.setSubscriptionKeyParameter(subscriptionKey)
+        if (this.api().type === TypeOfApi.webSocket || this.isGraphQL()) {
+            this.setSubscriptionKeyParameter(subscriptionKey);
```
malincrist, line 388: I see you moved the subscription-key from headers to parameters. But in the GraphQL console, the parameters are not shown as a separate section. Is this the expected behavior?

javierbo1989: This is something we haven't discussed before, but I think we definitely want to support this in the GraphQL console. All set, added. Thanks a ton for pointing this out.

src/components/operations/operation-details/ko/runtime/graphql-console.ts
```diff
@@ -681,6 +743,12 @@ export class GraphqlConsole {
 
     private editorValidations(): void {
         this.editorErrors([]);
+
+        const hasWsProtocol = !!this.api().protocols.find(p => p == "ws" || p == "wss");
+        if (this.isSubscriptionOperation() && !hasWsProtocol) {
+            this.editorErrors.push(`Websocket protocol not enabled, you can enable it in Azure > APIM > APIs > ${this.api().name} > Settings tab, and reload Developer Portal.`);
```
azaslonov, line 749: @javierbo1989, the dev portal is for C2 customers, they won't be able to go to Azure portal. How about something like "Live subscriptions are not supported for this API". @mikebudzynski can you please help with messaging?

src/components/operations/operation-details/ko/runtime/graphql-console.ts
```diff
@@ -600,18 +620,60 @@ export class GraphqlConsole {
...
+        const protocol = this.api().protocols.includes("wss") ? "wss" : "ws";
+        const urlTemplate = this.getRequestPath();
+        const result = `${protocol}://${this.host.hostname()}${Utils.ensureLeadingSlash(urlTemplate)}`;
+        return result;
+    }
+
+    private getRequestPath(getHidden: boolean = false): string {
+        let versionPath = "";
+
+        if (this.api().apiVersionSet && this.api().apiVersion && this.api().apiVersionSet.versioningScheme === "Segment") {
```
azaslonov, line 639: `this.api()` - gets evaluated multiple time in the same function. I'd suggest doing `const api = this.api()` and use that everywhere below.

src/components/operations/operation-details/ko/runtime/authorization.ts
```diff
@@ -410,15 +414,20 @@ export class Authorization {
...
+            this.updateRequestSummary();
+        }
     }
 
     private getSubscriptionKeyParam(): ConsoleParameter {
         const subscriptionKeyParamName = this.getSubscriptionKeyParamName();
         const searchName = subscriptionKeyParamName.toLocaleLowerCase();
-
-        return this.consoleOperation().request.queryParameters().find(x => x.name()?.toLocaleLowerCase() === searchName);
+        const queryParameters = (this.isGraphQL()) ? this.queryParameters() : this.consoleOperation().request.queryParameters();
```
azaslonov, line 429: `(this.isGraphQL())` - I guess no need to wrap in brackets.

src/components/operations/operation-details/ko/runtime/graphql-console.ts
```diff
@@ -600,18 +620,60 @@ export class GraphqlConsole {
...
+        let requestUrl = "";
+        const parameters = this.queryParameters();
+
+        parameters.forEach(parameter => {
+            if (parameter.value()) {
+                const parameterPlaceholder = parameter.name() !== "*" ? `{${parameter.name()}}` : "*";
+
+                if (requestUrl.indexOf(parameterPlaceholder) > -1) {
+                    requestUrl = requestUrl.replace(parameterPlaceholder,
+                        !getHidden || !parameter.secret ? Utils.encodeURICustomized(parameter.value())
```
azaslonov, line 652: `Utils.encodeURICustomized(parameter.value())` - repeated 3 times in the same function.

src/components/operations/operation-details/ko/runtime/authorization.ts
```diff
@@ -410,15 +414,20 @@ export class Authorization {
         keyParameter.required = true;
         keyParameter.inputType("password");
 
-        this.consoleOperation().request.queryParameters.push(keyParameter);
-        this.updateRequestSummary();
+        if (this.isGraphQL()) {
+            this.queryParameters.push(keyParameter);
```
azaslonov, line 418: `this.queryParameters.push(keyParameter);` and `this.consoleOperation().request.queryParameters.push(keyParameter);` - aren't they same?

src/components/operations/operation-details/ko/runtime/graphql-console.ts
```diff
@@ -331,6 +343,14 @@ export class GraphqlConsole {
         }
     }
 
+    private requestUrl(): string {
+        const protocol = this.api().protocols.includes("https") ? "https" : "http";
```
azaslonov, line 347: `includes("https") ? "https" : "http"` - begs for constant :)

azaslonov (commented):
Events like `connection_init`, `subscribe`, `next` would benefit from dedicated enum, something like `GqlWsMessageType`.

azaslonov (approved):

src/components/operations/operation-details/ko/runtime/authorization.ts
javierbo1989: Unfortunately no, consoleOperation is a class that we have specifically to handle REST, WebSocket, etc APIs. In this class are included the headers and parameters of the requests of these APIs. In our graphql solution we don't use this class, so we have to handle headers and parameters independently. 

For this reason, the graphql console passes these parameters independently 
image

and the operation console passes the instance of consoleOperation.
image

PR #1750, opened 2022-05-03
Added local styles for List of APIs dropdown widget

malincrist:


src/components/apis/list-of-apis/ko/runtime/api-list-dropdown.html
```diff
@@ -14,11 +14,11 @@
             <!-- /ko -->
             <!-- /ko -->
         </div>
-        <button class="input-group-addon no-border" data-toggle="dropdown" aria-label="Expand APIs list">
+        <button class="dropdown-input-button" data-toggle="dropdown" aria-label="Expand APIs list">
             <i class="icon-emb icon-emb-chevron-down"></i>
         </button>
-        <div class="dropdown">
-            <input type="search" class="form-control form-control-light" aria-label="Search" placeholder="Search APIs"
+        <div class="dropdown drop-down">
```
azaslonov, line 20: `drop-down` - maybe `dropdown-container` instead? Just looks a bit weird to have 2 "dropdown" classes here.

src/components/apis/list-of-apis/ko/listOfApisViewModelBinder.ts
```diff
@@ -30,11 +30,16 @@ export class ListOfApisViewModelBinder implements ViewModelBinder<ListOfApisMode
                 : undefined
         }));
 
+        const handler = model.layout == "list" ? ListOfApisHandlers :
+            model.layout == "dropdown" ?
+                ListOfApisDropdownHandlers :
+                ListOfApisTilesHandlers;
```
Mielek, line 36: Threat this comment as nit-pick and a great opportunity to learn.

In the next assignment you need to add a next type "popup" of handlers. Will you add a next case to that ternary operator?

```js
model.layout == "list" 
? ListOfApisHandlers 
: model.layout == "dropdown" 
?  ListOfApisDropdownHandlers 
: model.layout == "popup" 
? ListOfPopupHandlers
: ListOfApisTilesHandlers;
``` 
I think it is quite complex and had to maintain. Instead of that you can use object.

 ```js
const handlerForLayout = {
  list: ListOfApisHandlers,
  dropdown: ListOfApisDropdownHandlers,
  popup: ListOfPopupHandlers,
};
const handler = handlerForLayout[model.layout] ?? ListOfApisTilesHandlers;
```

I hope it will help you :)

Mielek (approved):

src/components/apis/list-of-apis/ko/listOfApisViewModelBinder.ts
malincrist: You are right. It is clearer like this and easier to maintain. Thanks for pointing it out.

PR #1768, opened 2022-05-16
Added local styles for API List tiles widget + refactoring

malincrist:


src/components/products/product-list/ko/productListViewModelBinder.ts
```diff
@@ -9,9 +9,16 @@ import { ProductListDropdownHandlers, ProductListHandlers, ProductListTilesHandl
...
+
     constructor(
         private readonly eventManager: EventManager,
-        private readonly styleCompiler: StyleCompiler) { }
+        private readonly styleCompiler: StyleCompiler) {
+            this.handlerForLayout = {
+                list: ProductListHandlers,
+                dropdown: ProductListDropdownHandlers
+            };
+         }
```
Mielek, line 21: Overall, this is good. I do not suspect that ProductListViewModelBinder is created thousands of times. It is really a minor.

The handlerForLayout field will be filled with new object every time class object is created. Two ProductListViewModelBinder object classes will have two equal instances handlerForLayout values which will reserve space in memory. 

To not use memory that you can use a static member to remove duplicating objects.
```ts
class SomeClass {
 private static readonly handlerForLayout = {
   list: ProductListHandlers,
   dropdown: ProductListDropdownHandlers,
 };
 someMethod() {
  ProductListViewModelBinder.handlerForLayout[layout]
 }
}
```

or you can just put a const variable outside the class block and use the value inside of the class

```ts
const handlers = { /* ... */ };
class SomeClass {
 someMethod() {
  handlers[layout]
 }
}
```

Mielek (approved):

src/components/products/product-list/ko/productListViewModelBinder.ts
malincrist: Even if it's not created thousands of times, there's no point in using extra memory when not needed. Good catch, thanks 😄

Mielek (approved):

PR #1774, opened 2022-05-18
Added local styles for User Subscription widget

malincrist:


src/components/users/subscriptions/subscriptionsHandlers.ts
```diff
@@ -14,4 +16,19 @@ export class SubscriptionsHandlers implements IWidgetHandler {
...
     }
+
+    public getStyleDefinitions(): StyleDefinition {
+        return {
+            components: {
+                userSubscriptions: {
+                    displayName: "User: Subscriptions",
+                    plugins: ["typography", "background", "margin", "padding"],
+                    components: {
+                        subscriptionsGridHeader: DefaultStyleDefinitions.GridHeader,
```
azaslonov, line 27: Is it possible to use `tableRow` name here to just override the base class instead of introducing `subscriptionsGridRow`?

Mielek (approved):

PR #1778, opened 2022-05-20
Local styles/api details

malincrist:


src/components/apis/details-of-api/ko/runtime/api-details.ts
```diff
@@ -94,6 +95,10 @@ export class ApiDetails {
         this.currentApiVersion(api.name);
         this.api(api);
 
+        this.api().contact ={ email: " f", name:"vejk", url:"nvwa"};
+        this.api().license = {name:"enwiovgw", url:"fnwiqoegneqio"}
+        this.api().termsOfServiceUrl= " fneigneow"
```
Mielek, line 100: is it not some leftover?

src/components/apis/list-of-apis/listOfApisHandlers.ts
```diff
@@ -68,27 +70,43 @@ export class ListOfApisTilesHandlers implements IWidgetHandler {
...
                     components: {
-                        searchInput: DefaultStyleDefinitions.SearchInput,
-                        apiCard: DefaultStyleDefinitions.Card,
-                        cardTitle: DefaultStyleDefinitions.CardTitle,
-                        cardText: DefaultStyleDefinitions.CardText,
-                        widgetText: DefaultStyleDefinitions.WidgetText,
-                        tagInput: DefaultStyleDefinitions.TagInput,
-                        toggleButtonLabel: DefaultStyleDefinitions.ToggleButtonLabel,
-                        tagCard: DefaultStyleDefinitions.TagCard
+                        searchInput: cloneDeep(DefaultStyleDefinitions.SearchInput),
```
Mielek, line 95: Can you describe why cloneDeep is needed?

src/components/defaultStyleDefinitions.ts
```diff
@@ -348,4 +348,95 @@ export const TagCard: ComponentStyleDefinition ={
...
+                width: "1",
+                style: "solid",
+                colorKey: "colors/borderColor"
+            },
+            right: {
+                width: "1",
+                style: "solid",
+                colorKey: "colors/borderColor"
+            }
+        },
```
Mielek, line 378: Why not just
```suggestion
        border: {
            width: "1",
            style: "solid",
            colorKey: "colors/borderColor"
        },
```

or

```suggestion
        borderWidth: "1",
        borderStyle:  "solid",
        borderColorKey: "colors/borderColor",
```


I think it should translate to something like
```scss
border-width: "1px";
border-style: solid;
border-color: $colors.borderColor;
```

src/components/apis/details-of-api/detailsOfApiContract.ts
```diff
@@ -1,9 +1,14 @@
...
 
 export interface DetailsOfApiContract extends Contract {
     /**
      * Link to a page that contains API changlog.
      */
     changeLogPageHyperlink?: HyperlinkContract;
+    /**
+     * Widget local styles.
+     */
+    styles: LocalStyles;
```
Mielek, line 13: Are styles here optional? if yes please do
```suggestion
    styles?: LocalStyles;
```

src/components/apis/details-of-api/detailsOfApiModel.ts
```diff
@@ -1,8 +1,14 @@
...
+export class DetailsOfApiModel {
     /**
      * Link to a page that contains API changelog.
      */
     public changeLogPageHyperlink: HyperlinkModel;
+
+    /**
+     * Widget local styles.
+     */
+    styles: LocalStyles;
```
Mielek, line 13: This is a class so if you require that field always need to have object under that field you can do the following

```suggestion
    styles: LocalStyles = {};
```

src/components/apis/details-of-api/detailsOfApiModelBinder.ts
```diff
@@ -7,14 +7,15 @@ import { IPermalinkResolver } from "@paperbits/common/permalinks";
...
-    
+
     public canHandleModel(model: Object): boolean {
         return model instanceof DetailsOfApiModel;
     }
 
     public async contractToModel(contract: DetailsOfApiContract): Promise<DetailsOfApiModel> {
         const model = new DetailsOfApiModel();
-
+        model.styles = contract.styles || {};
```
Mielek, line 17: if styles can be `undefined` then it is safer to use `??`

src/components/apis/details-of-api/ko/runtime/api-details.html
```diff
@@ -6,29 +6,29 @@
...
+            <span class="badge api-type-badge">GraphQL</span>
             <!-- /ko -->
         </h1>
 
         <div class="form-inline">
             <div class="form-group">
                 <!-- ko if: versionApis().length > 0 -->
                 <label for="apiVersions">API version: </label>
-                <select id="apiVersions" class="form-control"
+                <select id="apiVersions" class="api-versions-combobox"
```
Mielek, line 26: If it is possible, I would leave `form-control` class and added a modifier class, like yours `api-versions-combobox`

Mielek (changes requested):

src/components/apis/details-of-api/detailsOfApiHandlers.ts
```diff
@@ -14,4 +18,31 @@ export class DetailsOfApiHandlers implements IWidgetHandler {
 
         return widgetOrder;
     }
+    public getStyleDefinitions(): StyleDefinition {
+        return {
+            colors: {
+                badgeColor: {
+                    displayName: "API type badge color",
+                    defaults: {
+                        value: "#636363"
```
DeagleGross, line 27: are there any global color settings? Or it is a code convention to write down color-values explicitly in every component?

src/components/apis/list-of-apis/listOfApisHandlers.ts
malincrist: When modifying the style definitions for a widget, the object gets modified. So, without cloneDeep, all the widgets that have the same style definition will be modified:
clonedeep

src/components/defaultStyleDefinitions.ts
malincrist: Unfortunately, it is not possible.

When changing the border for a widget, users have to specify separate values for all the borders:
image

So the CSS class has to have different values for all of them, like:
```
border-top: 1px solid #505050;
border-left: 1px solid #505050;
border-bottom: 1px solid #505050;
border-right: 1px solid #505050;
```

to allow users to modify only the border they want (in case they want, for example, to have a different value for the bottom border only)

src/components/apis/details-of-api/detailsOfApiHandlers.ts
malincrist: There are global color settings. By defining the missing colors for components like this, they will be added dynamically in the global definitions.

The reason for this is that the global colors are defined in the default.json (which is the default stylesheet for the developer portal, before users make and save any customizations). 
Adding them there is not possible, since it would require users to reset their developer portal in order to get the new global stylesheet and have those changes.

DeagleGross: got it, thanks!

src/components/apis/details-of-api/ko/runtime/api-details.ts
```diff
@@ -6,6 +6,7 @@ import { ApiService } from "../../../../../services/apiService";
 import { Api } from "../../../../../models/api";
 import { RouteHelper } from "../../../../../routing/routeHelper";
 import { KnownMimeTypes } from "../../../../../models/knownMimeTypes";
+import { toHast } from "mdast-util-to-hast";
```
Mielek, line 9: Not used import

src/components/apis/details-of-api/ko/runtime/api-details.html
```diff
@@ -6,29 +6,29 @@
...
                 <!-- ko if: versionApis().length > 0 -->
                 <label for="apiVersions">API version: </label>
-                <select id="apiVersions" class="form-control"
+                <select id="apiVersions" class="form-control api-versions-combobox"
                     data-bind="options: versionApis, optionsValue: 'name', optionsText: 'apiVersion', value: currentApiVersion">
                 </select>
                 <!-- /ko -->
                 <!-- ko if: api().type !== 'websocket' && api().type !== 'graphql' -->
-                <select id="apiDefinitions" class="form-control" aria-label="API definitions"
+                <select id="apiDefinitions" class="api-definitions-combobox" aria-label="API definitions"
```
Mielek, line 31: fixed for line 26 but not for line 31. Is there a reason?

Mielek (approved):

PR #1785, opened 2022-05-26
Change workshop container height

Mielek:
Make workshop container scale with the size of window.

Here are some examples how workshop looks after this change
image
image
image

### DIFF
--- a/src/themes/designer/styles/workshops.scss
+++ b/src/themes/designer/styles/workshops.scss
@@ -40,8 +40,6 @@
         h1 {
             padding-left: 5px;
             line-height: 25px;
-
-         
         }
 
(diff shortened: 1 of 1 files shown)


malincrist:
What is the reason for this?
I think it look a bit weird to have so much free space under the toolbox buttons.

malincrist:
What is the reason for this?
I think it look a bit weird to have so much free space under the toolbox buttons.

malincrist:
@Mielek I see. Indeed, on smaller screens it looks good. And since it is needed for new features, I think you are right and it's better to have it as a percentage of the height rather than a fixed size. Thanks for the explanation.

malincrist (approved):

PR #1799, opened 2022-06-03
Dispatch errors to validation-summary refactor

JMach1:


src/components/users/validation-summary/utils.ts
```diff
@@ -0,0 +1,66 @@
+import { EventManager } from "@paperbits/common/events";
+import { ValidationReport } from "../../../contracts/validationReport";
+
+export const onValidationErrors = "onValidationErrors";
+
+export enum errorSources {
```
azaslonov, line 6: I guess we could move it to separate file. Also, please use capital case for enums.

src/components/users/validation-summary/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+        await guardedFunction();
+    } catch (error) {
+        return parseAndDispatchError(eventManager, source, error, fallbackMessage, errorDetailsMap);
+    }
+}
+
+export function parseAndDispatchError(
+    eventManager: EventManager,
+    source: errorSources,
+    error: Record<string, any>,
```
azaslonov, line 39: We have `MapiError` class, which MapiClient throws.

src/components/users/change-password/ko/runtime/change-password.ts
```diff
@@ -140,38 +131,8 @@ export class ChangePassword {
...
 
-            const validationReport: ValidationReport = {
-                source: "changepassword",
-                errors: []
-            };
-            this.eventManager.dispatchEvent("onValidationErrors", validationReport);
-        } catch (error) {
-            if (isCaptcha) {
-                await this.refreshCaptcha();
-            }
```
ygrik, line 152: this block is required for refresh the captcha widget on error.

src/components/users/profile/ko/runtime/profile.ts
```diff
@@ -128,31 +119,13 @@ export class Profile {
             return;
         }
         
-        this.cleanValidationErrors();
         this.working(true);
 
-        try {
+        await tryCatchDispatchError(async () => {
```
azaslonov, line 124: `await tryCatchDispatchError(async () => {` - I can't say I like this pattern. This maybe Ok for these cases, but if we want to handle other errors, not related to validation, this becomes problematic.

Maybe we could do something like this (keeping your new utils):

```
try {
   ...
}
catch (error) {
    if (error instanceof ValidationError) {
         handleValidationError(this.eventManager, errorSources.signInOAuth, error);
    }
    else if (error instanceof OtherError) {
       ...
    }
}
finally {
   ...
}
```

(we don't have ValidationError class, but MapiClient could throw one when the code is "ValidationError" instead of generic MapiError)

src/components/users/change-password/ko/runtime/change-password.ts
JMach1: I totally missed that, my bad, fixed. Great catch!

src/components/users/profile/ko/runtime/profile.ts
JMach1: I wasn't sure about this part either, why I included it in the end was only for several repetitive cases of 
```
dispatchErrors(eventManager, source, []); // clear error
try {
    foo();
  } catch (error) {
    // dispatch only the error and don't do anything else
    return parseAndDispatchError(eventManager, source, error, fallbackMessage, errorDetailsMap);
}
```
but yes, if requirement change and we need to do something in catch block, it might be confusing what to do so I removed the function so it's more verbose but clearer

JMach1: You can see nice example of what I originally had in mind in this commit: https://github.com/Azure/api-management-developer-portal/pull/1799/commits/748afedc34201942a7cc29d7e3c05f77abd69099

azaslonov (approved):

PR #1802, opened 2022-06-06
tslint --fix

JMach1:
I noticed there's a lot of TSLint errors, so I at least ran `tslint -p . --fix` to fix auto-fixable ones and created it as a npm script. Now before you push something, you can simply run `npm run lint:fix` to make sure, your code does not include any (auto-fixable) TSLint errors. (You should even be able to configure your IDE to do that for you. ;) )

If somebody does not like some of those changes, please, reconfigure the TSLint and commit changes to this branch. Or propose to remove TSLint completely if you would wish to, I just don't like having it included in the project and not following its warnings.

Adding as reviewers all, who I know regularly contribute to this repo to have an opportunity to share your thoughts. Thanks!


azaslonov (approved):

azaslonov:
Thanks Jan! I wonder if we can even add it to the gate.

PR #1811, opened 2022-06-08
Fixed height in graphql console

javierbo1989:


malincrist (commented):
Why not just set the background of the `<graphql-console>` (which has the height set to 100%) to the same grey as the divs inside?

javierbo1989:
Hi @malincrist, your solution is better, like it. Thank you very much.
All set, also added other minor changes that I noticed while doing this. Please, let me know any other comments.

src/components/operations/operation-details/ko/runtime/operation-details.html
```diff
@@ -338,14 +338,14 @@ <h2>GraphQL endpoint</h2>
...
-    <operation-console class="test flex flex-column gql-console-height" data-bind="dialog: {}"
+    <operation-console class="flex flex-column h-100" data-bind="dialog: {}"
         params="{ api: api, operation: operation, hostnames: hostnames, authorizationServer: associatedAuthServer, useCorsProxy: useCorsProxy }">
     </operation-console>
 </div>
 <!-- /ko -->
 <!-- ko if: apiType() === 'graphql' -->
 <div class="detachable-right graphql-console-width flex-grow animation-fade-in">
-    <graphql-console class="gql-console-height flex flex-column" data-bind="dialog: {}"
+    <graphql-console class="flex flex-column h-100 gql-console-bg" data-bind="dialog: {}"
```
malincrist, line 348: We have a style defined for operation-console that has the background and height:
image

Since we want the same background and height for `operation-console` and  `graphql-console`, we could create a class that has these properties and use it for both of them(and remove the operation-console style). 

Or, if you think it's better to have separate styles for them, maybe we can create a similar style for the `graphql-console`?

What do you think?

malincrist (approved):

src/components/operations/operation-details/ko/runtime/operation-details.html
javierbo1989: Both consoles seem to have some particular styles. But I think we could do something like this to define what would be common and what would be independent for both. Thanks.
image
image

azaslonov (approved):

PR #1821, opened 2022-06-15
Release date column should show date of creation of release

### LINKED ISSUE #1812: Release date column of API changelog shows updatedDateTime instead of createdDateTime
## Bug description

The release date column of an API Change Log entry shows the `updatedDateTime` instead of the `createdDateTime`.

## Reproduction steps

1. Creating an API and add a few releases over multiple days
2. Update an existing release log entry of an older date e.g. via Azure Resource Explorer
3. See that the developer portal change logs of the API show now the current date instead of the original time of the release.

## Expected behavior

As the column is named `Release Date` I would expect that the original date is kept here, because only the Notes of the release have been changed, not the content of the release.

We want to be able to update the release notes afterwards via ARM-Template, after an initial creation of a release entry, because:
- Maybe there was a typo/misspelled word in the release note, which needs to be corrected
- There was something forgotten in the release note

Managed

## Additional context

Add any other context about the problem here, including screenshots.
image
image

Line of code which takes the `updatedDateTime` instead of `createdDateTime`:
https://github.com/Azure/api-management-developer-portal/blob/e2f535bc7e907473b32dd0200b50437e0a5cfd5e/src/components/apis/history-of-api/ko/runtime/api-history.html#L28

dldl-cmd:
Release date column should show date of creation of release, not date of release notes update.

Fixes #1812

### DIFF
--- a/src/components/apis/history-of-api/ko/runtime/api-history.html
+++ b/src/components/apis/history-of-api/ko/runtime/api-history.html
@@ -25,7 +25,7 @@
                 <!-- ko if: currentPageLog().length > 0 -->
                 <div data-bind="foreach: { data: currentPageLog, as: 'hist' }">
                     <div class="table-row">
-                        <div class="col-3" data-bind="text: hist.properties.updatedDateTime"></div>
+                        <div class="col-3" data-bind="text: hist.properties.createdDateTime"></div>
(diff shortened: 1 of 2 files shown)


azaslonov (approved):

azaslonov:
@dldl-cmd, many thanks for your contribution!

PR #1826, opened 2022-06-16
Added local styles for confirm password widget

malincrist:


src/components/users/confirm-password/confirmPasswordModel.ts
```diff
@@ -1 +1,8 @@
-export class ConfirmPasswordModel { }
+import { LocalStyles } from "@paperbits/common/styles";
+
+export class ConfirmPasswordModel {
+    /**
```
azaslonov, line 4: Formatting of the jsdocs is a bit shifted. It should be like this:
```
    /**
     * Widget local styles.
     */
```
Just in case: https://jsdoc.app/about-getting-started.html

azaslonov (approved):

PR #1832, opened 2022-06-21
Added local styles for Reports widget

malincrist:


src/components/reports/reportsHandlers.ts
```diff
@@ -14,4 +16,34 @@ export class ReportsHandlers implements IWidgetHandler {
...
+                        value: "#dee2e6"
+                    }
+                }
+            },
+            components: {
+                reports: {
+                    displayName: "Reports",
+                    plugins: ["background", "margin", "padding"],
+                    components: {
+                        lastHourButton: DefaultStyleDefinitions.getDefaultButtonStyleDefinition(),
```
azaslonov, line 35: I guess no need to style every button individually, I'd say we can think of it as "tab" or "switch" control which has similar elements just with different labels on it and to states "normal" and "active".

azaslonov (approved):

src/components/reports/reportsHandlers.ts
malincrist: I think you are right. It makes more sense to look at it as a tab/switch control and use the same style for all the buttons. Thanks for pointing it out.

PR #1836, opened 2022-06-23
Fixed Auth header not being cleared on No auth select

jsorohova:


jsorohova:
Looks like someone forgot to push package-lock with package.json, so pushing as well.

azaslonov (approved):

PR #1839, opened 2022-06-23
Fixed operation list layout and accessibility issues.

azaslonov:


src/themes/website/styles/docs.scss
```diff
@@ -214,15 +212,13 @@ validation-summary {
     display: block;
 }
 
-.consoles {
-    background: $panel-bg-color-highlight;
-    height: 100%;
+operation-console {
```
malincrist, line 215: `.consoles` class was introduced to define the common styles from both `operation-console` and `graphql-console`. 
Also, if I remember correctly, removing it will re-introduce the bug with the height of the graphql-console, as the background for the console was defined here.

Why would we remove it now?

package.json
```diff
@@ -29,87 +29,87 @@
...
+    "rehype-sanitize": "^5.0.1",
+    "rehype-stringify": "^9.0.3",
+    "remark": "^14.0.2",
+    "remark-gfm": "^3.0.1",
+    "remark-parse": "^10.0.1",
+    "remark-rehype": "^10.1.0",
+    "saxen": "^8.1.2",
+    "topojson-client": "^3.1.0",
+    "truncate-html": "^1.0.4"
+  }
```
JMach1, line 114: Could you please take a look how/why is it changing the indentation of the "dependencies" to two spaces while all other keys are indented with four? Or are you doing this on purpose for some reason?
It's not huge issue, it's just producing unnecessary changes every time and I believe it should be indented the same way in whole file.

src/themes/website/styles/docs.scss
azaslonov: Good catch! That was messed up merge from managed repo. Fixed it.

src/components/operations/operation-list/ko/runtime/operation-list.html
```diff
@@ -93,23 +87,17 @@
...
-                    </div>
-                </a>
-            </li>
+            <div role="option" class="list-item table-preset-operations" data-bind="activate: $component.selectOperation, css: { 'selected': $component.selectedOperationName() === item.name }">
+                <div class="http-method" data-bind="attr: { 'data-method': method }"></div>
+                <div data-bind="text: ($component.showUrlPath() ? item.urlTemplate : item.displayName), css: { 'text-truncate': !$component.wrapText(), 'text-wrap': $component.wrapText() }, attr: { title: ($component.showUrlPath() ? item.urlTemplate : item.displayName) }"></div>
+            </div>
             <!-- /ko -->
-        </ul>
+            </div>
```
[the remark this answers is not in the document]
azaslonov: Applied formatting.

package.json
azaslonov: Applied formatting.

JMach1 (approved):

PR #1843, opened 2022-06-28
Fixed multiple mapcharts to work independently on mouse move

malincrist:
## Problem:
When hovering over a map chart, the specific country gets highlighted in all other maps:

## Solution:
Ensure every map has its own id and they work independently:

### Note:
Also fixed some issues/deprecations introduced by upgrading the d3 library


Hardell (approved):

DeagleGross:
best explanation of PR changes I have seen !

Hardell (approved):

malincrist:
@DeagleGross thank you! :)

PR #1848, opened 2022-06-30
Introduce base dev portal logger

f-alizada:


src/components/operations/operation-details/ko/runtime/operation-console.html
```diff
@@ -402,7 +402,7 @@ <h3>HTTP request
                     <i class="icon-emb icon-emb-eye"></i> Show
                     <!-- /ko -->
                 </button>
-                <button class="code-snippet-command" data-bind="copyToClipboard: $component.getPlainTextCodeSample"
+                <button class="code-snippet-command" data-bind="copyToClipboard: $component.getPlainTextCodeSample, click: function(data) {logCopyEvent();}"
```
azaslonov, line 405: `"click: function(data) {logCopyEvent();}"` can be just `"click: logCopyEvent"`.

src/components/operations/operation-details/ko/runtime/operation-console.ts
```diff
@@ -73,7 +74,8 @@ export class OperationConsole {
         private readonly tenantService: TenantService,
         private readonly httpClient: HttpClient,
         private readonly routeHelper: RouteHelper,
-        private readonly settingsProvider: ISettingsProvider
+        private readonly settingsProvider: ISettingsProvider,
+        private readonly logger: ConsoleLogger
```
azaslonov, line 78: This needs to be interface, not specific logger: `private readonly logger: Logger`. This way by changing the logger registration, we can use different loggers depending on environment.

src/models/console/consoleOperation.ts
```diff
@@ -11,6 +11,7 @@ import { ConsoleHeader } from "./consoleHeader";
 export class ConsoleOperation {
     private readonly api: Api;
 
+    public readonly id: string;
```
azaslonov, line 14: Do we really need this id? `public readonly id: string;`
@harunrst is moving the portal to new Data API, where there is no `id` property anymore. You folks may want to syncup about it.

src/components/operations/operation-details/ko/runtime/operation-console.ts
```diff
@@ -491,8 +493,11 @@ export class OperationConsole {
                     this.responseBody(Utils.formatXml(this.responseBody()));
                 }
             }
+
+            this.logSentRequest(this.api().id, consoleOperation.id, method, response.statusCode.toString());
```
azaslonov, line 497: One more case where `id` property will be gone: `this.api().id`. Please use `name` instead (not to be confused with `displayName`). Same goes to any other resource.

src/components/operations/operation-details/ko/runtime/operation-console.html
f-alizada: Thank you, updated!

src/components/operations/operation-details/ko/runtime/operation-console.ts
f-alizada: Got it, changed to interface

src/models/console/consoleOperation.ts
f-alizada: id of operation changed to the operationName which is introduced in ConsoleOperation model

src/components/operations/operation-details/ko/runtime/operation-console.ts
f-alizada: Update to name, will keep in mind that.

src/logging/devPortalLogger.ts
```diff
@@ -0,0 +1,81 @@
...
+        let devPortalEvent = new DevPortalEvent();
+        
+        devPortalEvent.eventType = eventName;
+        devPortalEvent.message = properties?.message;
+        devPortalEvent.eventData = JSON.stringify(properties);
+        
+        this.traceEvent(devPortalEvent);
+    }
+
+    public async trackError(error: Error, properties?: Bag<string>): Promise<void> {
```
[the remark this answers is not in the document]
f-alizada: trackError and trackView implementation were added, however they are not used in current functionality. Only trackEvent is being used.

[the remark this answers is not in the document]
azaslonov: They are not used in the dev portal code base itself, but they are used in the core libs. For instance, in the [error handler](https://github.com/paperbits/paperbits-common/blob/master/src/errors/unhandledErrorHandler.ts).

azaslonov (approved):

src/logging/devPortalLogger.ts
```diff
@@ -0,0 +1,93 @@
+import { Bag } from "@paperbits/common";
+import { Logger } from "@paperbits/common/logging";
+import { HttpClient, HttpRequest, HttpHeader } from "@paperbits/common/http";
+import { ISettingsProvider } from "@paperbits/common/configuration";
+import { DevPortalEvent } from "../models/logging/devPortalEvent";
+import { v4 as uuidv4 } from 'uuid';
+import * as Constants from "./../constants";
+import { version } from "./../../package.json";
```
azaslonov, line 8: I think we can't do that `import { version } from "./../../package.json";` because we make the dev configuration a part of application and that may have adverse effects on application bundling, tree-shaking. There is actually a special [webpack plugin](https://webpack.js.org/plugins/define-plugin/) for that purpose, so I'd recommend to use that.

azaslonov, line 8: I think we can't do that `import { version } from "./../../package.json";` because we make the dev configuration a part of application and that may have adverse effects on application bundling, tree-shaking. There is actually a special [webpack plugin](https://webpack.js.org/plugins/define-plugin/) for that purpose, so I'd recommend to use that.

azaslonov (changes requested):

src/logging/devPortalLogger.ts
f-alizada: Thank you for the explanation, updated the way of retrieving the version from package json using this plugin. Applied only to runtime.

azaslonov (approved):

PR #1852, opened 2022-07-05
Fixed delegation flow

### LINKED ISSUE #1792: Delegating product subscription workflow doesn't work with Self Hosted Portal
**Every bug report should have precise description and reproduction steps; console traces or source code references are appreciated.**

**For assistance requests, contact Azure support or submit a post [on Stack Overflow](http://aka.ms/apimso). We don't provide support through GitHub Issues. Feature requests can be raised [on the Azure Feedback Forum](https://aka.ms/apimwish).**

## Bug description

We are trying to delegate product subscriptions workflow with Self-hosted developer portal and facing issue with 404 page. Delegation is enabled for both product subscription and sign-in/sign-up authentication. 

## Reproduction steps

1.	Enabled delegation for both product subscription and sign-in/sign-up authentication as shown in delegation setting image.
2.	Login Developer PortalProductsTest-ProductSubscribe
3.	Redirect Call is [https://<serviceurl>/delegation-subscribe?productId=test-product&userId=<userid](https://%3Cserviceurl%3E/delegation-subscribe?productId=test-product&userId=%3cuserid)>

image

## Expected behavior

•	Subscription has been done successfully.
•	Redirect to User Profile page with Subscription details.

Actual Result : 
•	The Redirect Call that returns 404 is  [https://<serviceurl>/delegation-subscribe?productId=test-product&userId=<userid](https://%3Cserviceurl%3E/delegation-subscribe?productId=test-product&userId=%3cuserid)

image

 Self-hosted

Release 2.14.0

APIM

## Environment

 - Operating system: Windows
 - Browser: Chrome

## Additional context

Wok fine with legacy developer portal

ygrik:
fixed #1792

### DIFF
--- a/src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
+++ b/src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
@@ -157,9 +157,13 @@ export class ProductSubscribe {
                 const delegationParam = {};
                 delegationParam[DelegationParameters.ProductId] =  productName;
                 delegationParam[DelegationParameters.UserId] =  Utils.getResourceName("users", userId);
-                const delegationUrl = await this.backendService.applyDelegation(DelegationAction.subscribe, delegationParam);
+                const delegationUrl = await this.backendService.getDelegationString(DelegationAction.subscribe, delegationParam);
                 if (delegationUrl) {
-                    await this.router.navigateTo(delegationUrl);
+                    if (delegationUrl.startsWith("http://") || delegationUrl.startsWith("https://")){
+                        location.assign(delegationUrl);
+                    } else {
+                        await this.router.navigateTo(delegationUrl);
+                    }
                     return;
                 }
             }
--- a/src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
+++ b/src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
@@ -155,7 +155,7 @@ export class ProductSubscribe {
         try {
             if (this.delegationEnabled) {
                 const delegationParam = {};
-                delegationParam[DelegationParameters.ProductId] =  Utils.getResourceName("products", productId);
+                delegationParam[DelegationParameters.ProductId] =  productName;
                 delegationParam[DelegationParameters.UserId] =  Utils.getResourceName("users", userId);
                 const delegationUrl = await this.backendService.applyDelegation(DelegationAction.subscribe, delegationParam);
                 if (delegationUrl) {
--- a/src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
+++ b/src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
@@ -8,8 +8,9 @@ import { ProductService } from "../../../../../services/productService";
 import { UsersService } from "../../../../../services/usersService";
 import { SubscriptionState } from "../../../../../contracts/subscription";
 import { TenantService } from "../../../../../services/tenantService";
-import { DelegationParameters, DelegationActionPath } from "../../../../../contracts/tenantSettings";
+import { DelegationParameters, DelegationAction } from "../../../../../contracts/tenantSettings";
 import { RouteHelper } from "../../../../../routing/routeHelper";
+import { BackendService } from "../../../../../services/backendService";
 
 @RuntimeComponent({
     selector: "product-subscribe-runtime"
@@ -35,6 +36,7 @@ export class ProductSubscribe {
     constructor(
         private readonly usersService: UsersService,
         private readonly tenantService: TenantService,
+        private readonly backendService: BackendService,
         private readonly productService: ProductService,
         private readonly router: Router,
         private readonly routeHelper: RouteHelper
@@ -152,11 +154,14 @@ export class ProductSubscribe {
 
         try {
             if (this.delegationEnabled) {
-                const delegation = new URLSearchParams();
-                delegation.append(DelegationParameters.ProductId, Utils.getResourceName("products", productId));
-                delegation.append(DelegationParameters.UserId, Utils.getResourceName("users", userId));
-                await this.router.navigateTo(`/${DelegationActionPath.subscribe}?${delegation.toString()}`);
-                return;
+                const delegationParam = {};
+                delegationParam[DelegationParameters.ProductId] =  Utils.getResourceName("products", productId);
+                delegationParam[DelegationParameters.UserId] =  Utils.getResourceName("users", userId);
+                const delegationUrl = await this.backendService.applyDelegation(DelegationAction.subscribe, delegationParam);
+                if (delegationUrl) {
+                    await this.router.navigateTo(delegationUrl);
+                    return;
+                }
             }
 
             if (!this.subscriptionName()) {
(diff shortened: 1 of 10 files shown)


src/services/backendService.ts
```diff
@@ -144,6 +144,23 @@ export class BackendService {
         throw new MapiError("Unhandled", "Unable to complete change password request.");
     }
 
+    public async applyDelegation(action: DelegationAction, delegationParameters: {}): Promise<string> {
```
azaslonov, line 147: Could be `Bag<string>`.

src/components/users/subscriptions/ko/runtime/subscriptions.ts
```diff
@@ -140,10 +142,12 @@ export class Subscriptions {
...
         if (isDelegationEnabled) {
-            const delegation = new URLSearchParams();
-            delegation.append(DelegationParameters.SubscriptionId, Utils.getResourceName("subscriptions", subscriptionId));
-            this.router.navigateTo(`/${DelegationActionPath.unsubscribe}?${delegation.toString()}`);
-
+            const delegationParam = {};
+            delegationParam[DelegationParameters.SubscriptionId] =  Utils.getResourceName("subscriptions", subscriptionId);
+            const delegationUrl = await this.backendService.applyDelegation(DelegationAction.unsubscribe, delegationParam);
+            if (delegationUrl) {
+                await this.router.navigateTo(delegationUrl);
```
azaslonov, line 149: Looks like you're doing redirect inside operation called `isDelegationEnabled`, which from its name sounds like it should only check for delegation. I suspect this will bring more confusion/bugs down the road. Maybe do redirect outside of after the check?

src/components/users/signup/ko/runtime/signup.ts
```diff
@@ -109,6 +109,12 @@ export class Signup {
                     } else {
                         this.usersService.navigateToHome();
                     }
+                } else {
+                    const redirectUrl = this.delegationUrl();
+
+                    if (redirectUrl) {
+                        location.assign(redirectUrl);
```
azaslonov, line 116: I think we should user a `Router` instead, because it can be mocked for unit testing, have custom logic for redirects/routing, etc.

src/components/products/product-subscribe/ko/runtime/product-subscribe.ts
```diff
@@ -152,11 +154,14 @@ export class ProductSubscribe {
 
         try {
             if (this.delegationEnabled) {
-                const delegation = new URLSearchParams();
-                delegation.append(DelegationParameters.ProductId, Utils.getResourceName("products", productId));
-                delegation.append(DelegationParameters.UserId, Utils.getResourceName("users", userId));
-                await this.router.navigateTo(`/${DelegationActionPath.subscribe}?${delegation.toString()}`);
-                return;
+                const delegationParam = {};
+                delegationParam[DelegationParameters.ProductId] =  Utils.getResourceName("products", productId);
```
azaslonov, line 158: In `delegationParam[DelegationParameters.ProductId] = Utils.getResourceName("products", productId);` can we use `productName` directly, without wrapping it into `productId` and extracting it back with `Utils.getResourceName`?

src/components/users/subscriptions/ko/runtime/subscriptions.ts
ygrik: Renamed function to `applyDelegation` to better understand what it is doing.
If we do redirect outside, we will repeatedly copy redirect code in many places.

src/services/backendService.ts
```diff
@@ -144,7 +145,24 @@ export class BackendService {
...
+            const queryParams = new URLSearchParams();
+            Object.keys(delegationParameters).map(key => {
+                const val = delegationParameters[key];
+                queryParams.append(key, val);
+            });
+            return `/${DelegationActionPath[action]}?${queryParams.toString()}`;
+        } else {
+            const delegationUrl = await this.getDelegationUrl(action, delegationParameters);
+            if (delegationUrl) {
+                location.assign(delegationUrl);
```
azaslonov, line 159: I believe we can have a single place where we compose the delegation URL (I suggest in `getDelegationUrl` of `backendService`), call it from the runtime components and do redirects in them respectively (using router everywhere, instead of location.assign). Maybe even do encapsulation like this:

subscription.ts
const url = backendService.getSubscriptionDelegationUrl(subscriptionName)
router.navigateTo(url);

signin.ts:
const url = backendService.getSigninDelegationUrl(subscriptionName)
router.navigateTo(url);

src/components/users/subscriptions/ko/runtime/subscriptions.ts
```diff
@@ -116,8 +118,8 @@ export class Subscriptions {
         subscription.isSRegenerating(true);
 
         try {
-            const isDelegationEnabled = await this.isDelegationEnabled(subscriptionId);
-            if (isDelegationEnabled) {
+            const isDelegationApplied = await this.applyDelegation(subscriptionId);
```
azaslonov, line 121: Since you have the redirect inside `applyDelegation` method, there is no point to have this `if (isDelegationApplied){}` statement. If we get to this point, it means no delegation happened. Therefore, `applyDelegation` method could return `void` for simplicity.

src/routing/signOutRouteGuard.ts
```diff
@@ -34,9 +34,13 @@ export class SignOutRouteGuard implements RouteGuard {
                         const identity = await this.mapiClient.get<Identity>("/identity", [await this.mapiClient.getPortalHeader("delegationSignOut")]);
 
                         if (identity) {
-                            const redirectUrl = await this.backendService.getDelegationUrl(DelegationAction.signOut, { userId: identity.id });
+                            const delegationParam = {};
```
azaslonov, line 37: In the other comment I'm suggesting to hide the composition..
```
const delegationParam = {};
delegationParam[DelegationParameters.UserId] =  identity.id;
const redirectUrl = await this.backendService.applyDelegation(DelegationAction.signOut, delegationParam);
```
inside specifically named methods, like `getSignoutDelegationUrl` or something like that.

src/routing/signOutRouteGuard.ts
```diff
@@ -34,9 +34,13 @@ export class SignOutRouteGuard implements RouteGuard {
                         const identity = await this.mapiClient.get<Identity>("/identity", [await this.mapiClient.getPortalHeader("delegationSignOut")]);
```
azaslonov, line 34: The `usersService` has `getCurrentUserId()` method. You could re-use it here.

src/components/users/signin/ko/runtime/signin.ts
```diff
@@ -81,7 +81,7 @@ export class Signin {
                 const redirectUrl = this.delegationUrl();
 
                 if (redirectUrl) {
-                    window.open(redirectUrl, "_self");
+                    location.assign(redirectUrl);
```
azaslonov, line 84: One more `location.assign`.

src/routing/signOutRouteGuard.ts
ygrik: If we use UserService we will lost info about delegation user signin.

azaslonov (approved):

PR #1858, opened 2022-07-14
Switching to the latest API version

maciejtreder:


scripts.v2/utils.js
```diff
@@ -47,7 +47,7 @@ async function getStorageSasTokenOrThrow(managementApiEndpoint, managementApiAcc
  * @param {string} managementApiAccessToken the SAS token
  */
 async function getStorageSasToken(managementApiEndpoint, managementApiAccessToken) {
-    const response = await request("POST", `https://${managementApiEndpoint}/subscriptions/00000/resourceGroups/00000/providers/Microsoft.ApiManagement/service/00000/portalSettings/mediaContent/listSecrets?api-version=2019-12-01`, managementApiAccessToken);
+    const response = await request("POST", `https://${managementApiEndpoint}/subscriptions/00000/resourceGroups/00000/providers/Microsoft.ApiManagement/service/00000/portalSettings/mediaContent/listSecrets?api-version=2021-08-01`, managementApiAccessToken);
```
f-alizada, line 50: Is it possible to make this value as a constant variable and include it in each request? So next time it will not be needed to change different files.

azaslonov: +1, I should have done this in the first place.

maciejtreder: @f-alizada Changed. Thanks for pointing this out.

azaslonov (approved):

PR #1957, opened 2022-09-27
Handle multiple authorization servers

malincrist:
This PR introduces displaying the authorization servers specified for an API:

The servers displayed are only the ones that have the property "useInApiDocumentation" set to true.

The properties displayed for the servers in the operation details are the following:
- Name (only if multiple servers are specified) 
- Description 
- Supported grant flows 
- Authorization endpoint URL 
- Token endpoint URL 
- Scopes 

Here is an example of how multiple authorization servers are displayed:
aaaalign

The servers used for the test console are the ones that have the property "useInTestConsole" set to true:
In the test console, there is a new selector, where the authorization server can be selected. 
The authentication flow remains the same: when the selected grant type is changed, the authentication flow is initiated.
multiple-auth-test-console

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/operation-details.ts
+++ b/src/components/operations/operation-details/ko/runtime/operation-details.ts
@@ -22,7 +22,6 @@ import {
 import { OAuthService } from "../../../../../services/oauthService";
 import { LruCache } from "@paperbits/common/caching/lruCache";
 
-
 @RuntimeComponent({
     selector: "operation-details"
 })
@@ -46,7 +45,8 @@ export class OperationDetails {
     public readonly sampleHostname: ko.Observable<string>;
     public readonly hostnames: ko.Observable<string[]>;
     public readonly working: ko.Observable<boolean>;
-    public readonly associatedAuthServer: ko.Observable<AuthorizationServer>;
+    public readonly apiDocumentationAuthServers: ko.Observable<AuthorizationServer[]>;
+    public readonly testConsoleAuthServers: ko.Observable<AuthorizationServer[]>;
     public readonly apiType: ko.Observable<string>;
     public readonly protocol: ko.Computed<string>;
 
@@ -59,7 +59,8 @@ export class OperationDetails {
         this.working = ko.observable(false);
         this.sampleHostname = ko.observable();
         this.hostnames = ko.observable();
-        this.associatedAuthServer = ko.observable();
+        this.apiDocumentationAuthServers = ko.observable();
+        this.testConsoleAuthServers = ko.observable();
         this.api = ko.observable();
         this.schemas = ko.observableArray([]);
         this.tags = ko.observableArray([]);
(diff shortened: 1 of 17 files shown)


src/constants.ts
```diff
@@ -293,7 +293,7 @@ export enum GrantTypes {
     password = "password"
 }
 
-export const managementApiVersion = "2021-04-01-preview";
+export const managementApiVersion = "2022-04-01-preview";
```
harunrst, line 296: @azaslonov is it fine to upgrade this?

src/services/oauthService.ts
```diff
@@ -40,43 +42,14 @@ export class OAuthService {
...
-            let authServer: AuthorizationServer;
-
-            authServer = await this.getOauth2ServerByApiId(apiId);
-
-            if (!authServer) {
-                authServer = await this.getOpenidconnectServerByApiId(apiId);
-            }
-
-            return authServer ?? undefined;
+            const authServers = await this.apiClient.get<Page<AuthorizationServerForClient>>(`apis/${apiId}/authServers/openidconnect`, [await this.apiClient.getPortalHeader("getAuthorizationServer"), Utils.getIsUserResourceHeader()]);
```
harunrst, line 47: you are only fetching openidconnect servers, arent we supposed to fetch oauth2 servers as well :) also, isn't there any use case where we fetch only a single server?

harunrst:
@mikebudzynski Hey Mike, can you check the ui/ux here, does it look fine?

src/services/oauthService.ts
harunrst: hmm, I see, but why only Oauth2 is coming together with API? Why dont we have openid with it? can you take a look at why it is like this? Also I am not sure if it is the case but, in case oauth2 settings are empty, you continue as it is instead of fetching the oauth2 server.

image

I think between those, you can fetch oauth2 if it is not returned with api to make sure.

malincrist: As discussed, I made the change to also make the call for OAuth2 servers.

mikebudzynski:
> @mikebudzynski Hey Mike, can you check the ui/ux here, does it look fine?

Could we align the labels and text inputs to the middle vertically? Other than that, it looks alright to me.

src/constants.ts
azaslonov: Yes, feel free to upgrade.

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -194,7 +194,7 @@ export class OperationDetails {
 
         let associatedAuthServers: AuthorizationServer[];
         if (api.authenticationSettings?.oAuth2AuthenticationSettings.length > 0) {
-            associatedAuthServers = api.authenticationSettings?.oAuth2AuthenticationSettings.map(x => new AuthorizationServer(x.authorizationServer));
+            associatedAuthServers = await this.oauthService.getOauthServers(api.id);
```
harunrst, line 197: there is a small issue here, lets say you have both kind of auth servers. openid and oauth2, on the line 197 and 201, you are assigning to the associatedAuthServers arrays. So you are overriding the previous values. at the end, you will only have openid servers because you fetch it the latest.

Instead of overriding, you should append the new values.
associatedAuthServers.push(...oauth2Servers)
associatedAuthServers.push(...openidServers)

malincrist: You can't have both of them (there are validations for this on the backend).
I changed the second one to `if else` since there is no point to check if we have openID if we already have oAuth2, thanks for noticing.

malincrist:
@mikebudzynski I have aligned the text vertically and updated the screenshot from the description with how it looks now. Could you please check if it looks ok to you now?

harunrst (approved):

oskrabanek:
This is great news. How can we add more Auth Servers to the API in API Management in Azure Portal? Currently, there is a dropdown allowing to add only one.

malincrist:
Hello @oskrabanek , thanks for reaching out! This feature is not yet released (neither in Developer Portal, neither in Azure Portal).

It is on our roadmap and, once it will be released, you will be able to add multiple Auth Servers from Azure Portal and they will be reflected in the Developer Portal, as you can see in the description of this PR. 😊

cermakp:
Hello @malincrist,

Could it be possible to share with us the roadmap, including the ETA for that feature? This feature will significantly simplify our workflow.

Thank you.

cermakp:
Hi, could you please provide ETA with roadmap? Thank you

PR #1996, opened 2022-11-09
Juss wes/style reference fix

juss-wes:
Added html-webpack-link-type plugin for proper CSS MIME type handling - fixing issues with Chrome usage

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
                 "graphql": "^15.5.0",
                 "graphql-config": "^4.0.1",
                 "graphql-language-service": "^3.1.4",
+                "html-webpack-link-type-plugin": "^1.1.1",
                 "idb-keyval": "^6.2.0",
                 "js-beautify": "^1.14.7",
                 "knockout": "^3.5.1",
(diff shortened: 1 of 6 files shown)


juss-wes:
@microsoft-github-policy-service agree

azaslonov:
@juss-wes do you mind giving more context on what issues this fixes? With examples, if possible. Thanks in advance!

juss-wes:
@azaslonov - Sure!  When I deployed the developer portal today, it did not work in Chrome 107.0.5304.88.  In Chrome, it complains that the CSS styles are tagged with text/html instead of text/css and blocks them from being used.  Edge and Firefox aren't as picky and DO allow those styles to be used.

MIME-Error

Developer-Portal-Error

This plugin helps to render those with the 'text/css' MIME type, to make sure they don't come across incorrectly.

azaslonov:
Looking at the URL it looks like the problem is on the server side which serves that styles.css file. It should set the right content-type in response. I have a suspicion that this file contained the error response rather than actual CSS. Therefore, I'd like to ask you to send your service name to apimportalfeedback@microsoft.com.

Unfortunately, in case of managed portal your pull request won't help here. Managed portal has completely different build and assembly pipelines.

juss-wes:
Ah, I see.  So this is just for the self-hosted portal.  Would this still help those using that type of developer portal?

I see some errors of missing images and things too, so maybe you're right.  Maybe the user can't get to the content and so it's returning a 404 page or something instead of actual styles.  I'll send information to that email.

juss-wes:
And oddly enough....it's working today.

mikebudzynski:
@juss-wes, thanks for opening this pull request!

@juss-wes and @azaslonov, since the issue is already resolved, can we close this PR or does it contain another improvement?

azaslonov:
@juss-wes, I'm closing this PR, since it's managed portal backend issue (we're looking into it).

juss-wes:
No worries.  Thanks!

PR #2012, opened 2022-11-21
Examples column added to operation details table

JMach1:


src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -329,6 +334,40 @@ export class OperationDetails {
...
+                let valueObj;
+                try {
+                    valueObj = JSON.parse(value);
+                } catch (e) {
+                    return;
+                }
+
+                const contentTypeObj = {}
+                Object.entries(valueObj).forEach(([key, val]) => {
+                    if (typeof val === 'object') return
```
harunrst, line 356: his return will break the for loop, lets say your first element is not object type but the second one is object, the second one will not be seen

JMach1: No, it only stops the current iteration:
image

PR #2021, opened 2022-11-25
Fixed: Sign in button visible after successful sign in

malincrist:


src/components/users/signin-social/signinSocialModelBinder.ts
```diff
@@ -16,7 +16,15 @@ export class SigninSocialModelBinder implements IModelBinder<SigninSocialModel>
     public async contractToModel(contract: SigninSocialContract): Promise<SigninSocialModel> {
```
harunrst, line 16: about contract.roles, as i see you get it: map it under security and assign null. it seems like we dont need it, why dont we fill security from the beginning instead of roles? so that we wouldnt need any mapping and extra property

src/components/users/signin-social/signinSocialModelBinder.ts
```diff
@@ -16,7 +16,15 @@ export class SigninSocialModelBinder implements IModelBinder<SigninSocialModel>
...
-        model.roles = contract.roles || [BuiltInRoles.everyone.key];
+        if (contract.roles) { // migration.
+            contract.security = {
+                roles: contract.roles
+            }
+
+            contract.roles = null;
+        }
+
+        model.security = contract.security ?? { roles: [BuiltInRoles.everyone.key] };
```
harunrst, line 27: { roles: [BuiltInRoles.everyone.key] } is duplicated below on line 49 as well, we can const it as initialSecurity somewhere above and use it.

src/components/users/signin-social/signinSocialModelBinder.ts
malincrist: "security" is the new property that is used for the visibility based on roles. Previously, it used to be "roles", so after we merge this, there will be contracts that have the roles mapped under "roles" property (and nothing for the "security", so the widget will contain the default value).
This is why I mapped them to security, so we migrate the old customer data(roles) to the new model(security).

We do not keep the "roles" property anymore, if you take a look at `modelToContract`, we don't save it. So I guess assigning null after mapping is useless, I removed it, thanks for pointing it out.

src/components/users/signin-social/signinSocialModelBinder.ts
malincrist: done

JMach1 (approved):

PR #2034, opened 2022-12-09
Implemented the display of security schemes as they are in OpenAPI

malincrist:
Below is an example of how the Security Schemes are displayed:

image

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/apiKeyDetails.ts
+++ b/src/components/operations/operation-details/ko/runtime/apiKeyDetails.ts
@@ -1,0 +1,9 @@
+export enum ApiKeyLocation {
+    Query = "query",
+    Header = "header"
+}
+
+export interface ApiKeyDetails {
+    in: string;
+    name: string;
+}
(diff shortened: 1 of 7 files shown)


src/components/operations/operation-details/ko/runtime/apiKeyDetails.ts
```diff
@@ -0,0 +1,9 @@
+export enum ApiKeyLocation {
+    Query = "query",
+    Header = "header"
+}
+
+export interface ApiKeyDetails {
```
harunrst, line 6: Instead of ApiKey, AccessToken would be more accurate I think. Because It refers to here I guess:
image

malincrist: I used ApiKey to respect the OpenAPI specifications: https://swagger.io/docs/specification/authentication/
Since the idea of this feature is to display the security schemes as they are in OpenAPI, I think it makes more sense to use the same naming convention as they do.

I guess AccessToken make sense, but it will still be mapped to `API Key` in the Security Schemes section, which, I think, would be confusing. 

What do you think?

harunrst: yes it makes sense then, thanks!

harunrst (approved):

PR #2050, opened 2022-12-27
Custom widgets iframes media & clipboard access allowed

JMach1:


src/components/custom-widget/ko/runtime/customWidget.html
```diff
@@ -1,4 +1,5 @@
 <iframe data-bind="attr: { id: instanceId, name: name, src: src, sandbox: iframeSandboxAllows }"
+        allow="clipboard-read; clipboard-write; camera; microphone"
```
f-alizada, line 2: Is there a possibility to move the value of the permissions to one constant and propogate it to the required elements? 
decreases amount of code duplicates. 
What do you think?

JMach1: I thought about that, to do it same as the `sandbox: iframeSandboxAllows` is done. But iframeSandboxAllows is used on more places and is way longer. Meanwhile allow attribute makes sense that it might be different on various iframes. Also because of knockout, reusability of a constant is quite annoying to do.
But if you insists, I can change it.

f-alizada: From the PR it seems it will be used for customWidget for now and relates only to this. I believe we can update the PR to use constant, in case of any changes will be made we can change only one variable without modifying three files. 
Feel free to close the comment if you think it is not a big difference. 
Thank you

PR #2061, opened 2023-01-11
E2E tests

JMach1:
Add infra for running e2e tests. 

- Mock server per each test => flexibility for different cases
- pipeline
- examples of e2e tests


package.json
```diff
@@ -18,10 +18,12 @@
         "build-runtime": "webpack --config webpack.runtime.js",
         "build-function": "webpack --config webpack.function.js",
         "test": "node node_modules/mocha/bin/_mocha -r mocha.js src/**/*.spec.ts tests/e2e/**/*.spec.ts --timeout 30000",
+        "test-e2e": "node node_modules/mocha/bin/_mocha -r mocha.js tests/e2e/**/*.spec.ts --timeout 3000000",
```
harunrst, line 21: @f-alizada you might want to sync up with @JMach1 
he is running a specific test folder for e2e. And your branch is running all tests, we are double running tests.

f-alizada: This PR is on hold currently, and from change we can see that for e2e tests we have separate folder so both pr's running different tests

harunrst: hmm yes, it is under different folder than src, okay it is cool here then

DeagleGross (approved):
LGTM, but suggesting to wait for devPortal expert to review

.github/workflows/test.yml
```diff
@@ -18,3 +18,27 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Install
+        run: npm install
+
+      - name: Build static data
+        run: npm run build-mock-static-data
+
+      - name: Start mock server
+        run: npm run serve-website &
```
Mielek, line 36: is the & at the end needed?

malincrist (approved):

tests/e2e/runtime/apis.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        const page = await browser.newPage();
+        await page.goto(config.urls.apis);
+
+        const apiWidget = new ApisWidget(page);
+        await apiWidget.apis();
+
+        expect(await page.evaluate(() =>
+            document.querySelector("api-list div.table div.table-body div.table-row")?.parentElement?.childElementCount
+        )).to.equal(apis.apiList.length);
```
Mielek, line 37: I would suggest as you have a great opportunity to create a great example of test. In my opinion two elements are missing.

1.  Abstraction to set backend data - Abstraction which would make creating data easy and test very readable.
2.  Object oriented representation of page and page elements - This is for future. As tests will get bigger and more complex parts will be reused in many tests. It will be huge time saver in the future.

Please compare the following snipped with current test code. 

```typescript
if("User can see apis on the page", async () => {
  var backendData = Banked.createApi(TestApis.Create());
  
  const portal: DeveloperPortal = await DeveloperPortal.create(browser);
  const apisPage: ApisPage = await portal.openApisPage();
  const apisWidget: ApisWidget = await apisPage.getApisWidget();
  expect(await apisWidget.getListLength()).to.equal(backendData.apis.length);
  // or just asking ApisPage for most common elements/operations
  expect(await apisPage.getApiListLenght()).to.equal(backendData.apis.length);
});
```

.github/workflows/test.yml
f-alizada: Yes to run it in the background and not to wait to complete.

tests/e2e/runtime/apis.spec.ts
f-alizada: Thank you @Mielek. Totally agree with both suggestions, and the example looks great, do you mind if it will be done after this pr gets merged so we could populate by tests + get more feedback on how it is and implement the suggestions all at once as an improvement?

Mielek: I am ok with that, just want to make sure that it is not lost and it will be done. This is a big opportunity to have a great guideline how to properly do such tests.

Mielek (approved):

PR #2064, opened 2023-01-12
Add unit test workflow

f-alizada:
Add unit test workflow to run against pr to main branch.


.github/workflows/unitest.yml
```diff
@@ -0,0 +1,21 @@
+name: Unit test
```
harunrst, line 1: it doesn't specifically seem like unit tests. we are running all test files in this workflow, let's rename it as just Test or specify unit tests with its folder or however we store it

f-alizada: It runs all .spec files in the src folder, for e2e test I believe there is different folder, having that it runs unit test? Or Am I missing something?

f-alizada: Please Feel free to reopen the comment if I missed something, 
Thank you

f-alizada: Got it, thanks for review, updated

harunrst (approved):

Issue #2052, opened 2022-12-28
Expose the leave account button when using delegation


ika-microsoft:
Today, if you use delegation, the API portal does not show the button that invokes the leave account delegation.

GustavoFranzolin:
Do you have plans to make this feature available?

PR #2073, opened 2023-01-20
Display close account button for users signed in with delegation

### LINKED ISSUE #2042: Cannot close account when using Azure AD or Azure AD B2C Identity Providers
## Bug description

Cannot close account when using Azure AD or Azure AD B2C Identity Providers

## Reproduction steps

1 Configure the Developer portal with Azure AD and/or Azure AD Identity Providers
2. Log in to the Developer Portal with an Azure AD or Azure AD B2C account
3. Navigate to the user profile page
4. Notice that profile data and subscription information is shown but the button to close the account is missing 

## Expected behavior

Regardless of what Identity Provider is used a Developer Portal user should be able to close her/his account. 

Managed 

[e.g., contoso-api]

## Environment

 - Operating system: [e.g., iOS]
 - Browser: [e.g., Google Chrome, Safari]
 - Version: [e.g., 22]

## Additional context

It seems that the display of the close account-button is conditional, as can be seen here: 

https://github.com/Azure/api-management-developer-portal/blob/master/src/components/users/profile/ko/runtime/profile.html

malincrist:
Closes #2042 , #2052

### DIFF
--- a/src/components/users/profile/ko/runtime/profile.ts
+++ b/src/components/users/profile/ko/runtime/profile.ts
@@ -143,7 +143,9 @@ export class Profile {
         if (isDelegationApplied) {
             return;
         }
-        const confirmed = window.confirm(`Dear ${this.user().firstName} ${this.user().lastName}, \nYou are about to close your account associated with email address ${this.user().email}.\nYou will not be able to sign in to or restore your closed account. Are you sure you want to close your account?`);
+        const confirmed = window.confirm(
+            this.isBasicAccount() ? `Dear ${this.user().firstName} ${this.user().lastName}, \nYou are about to close your account associated with email address ${this.user().email}.\nYou will not be able to sign in to or restore your closed account. Are you sure you want to close your account?`
+            : `Dear ${this.user().firstName} ${this.user().lastName}, \nYou are about to close your account associated with email address ${this.user().email}.\nAre you sure you want to close your account?`);
 
         if (confirmed) {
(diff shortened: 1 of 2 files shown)


src/components/users/profile/ko/runtime/profile.ts
```diff
@@ -143,7 +143,9 @@ export class Profile {
         if (isDelegationApplied) {
             return;
         }
-        const confirmed = window.confirm(`Dear ${this.user().firstName} ${this.user().lastName}, \nYou are about to close your account associated with email address ${this.user().email}.\nYou will not be able to sign in to or restore your closed account. Are you sure you want to close your account?`);
+        const confirmed = window.confirm(
+            this.isBasicAccount() ? `Dear ${this.user().firstName} ${this.user().lastName}, \nYou are about to close your account associated with email address ${this.user().email}.\nYou will not be able to sign in to or restore your closed account. Are you sure you want to close your account?`
+            : `Dear ${this.user().firstName} ${this.user().lastName}, \nYou are about to close your account associated with email address ${this.user().email}.\nAre you sure you want to close your account?`);
```
f-alizada, line 148: Is there a possibility to move this text to some contants? 
It will be easier to mdify,change in future, and not change html file itself,
What do you think?

malincrist: as discussed, I extracted them in methods, thanks :)

jsorohova (approved):

Issue #2112, opened 2023-02-21
If you have an API named "operations" - operation name resolution fails and you are unable to retrieve the operations


cristiteodorov:
**Every bug report should have precise description and reproduction steps; console traces or source code references are appreciated.**

**For assistance requests, contact Azure support or submit a post [on Stack Overflow](http://aka.ms/apimso). We don't provide support through GitHub Issues. Feature requests can be raised [on the Azure Feedback Forum](https://aka.ms/apimwish).**

## Bug description

If you have an API named "operations", you cannot retrieve the operations of that API in the developer portal.

![image](https://user-images.githubusercontent.com/31003712/220315519-a4b579bb-6911-4f52-893e-94cafb014627.png)

This is due to the name for the operations being wrong.

## Reproduction steps

1. Create an API with "Name" set to "operations" in APIM. (the non editable "Name" field must be set to operations, not the "Display Name")
2. Point the API developer portal to that APIM instance.
3. Try to access an operation within the API named "operations" - you will notice an error in console: "Error: Resource not found..."

## Expected behavior

When selecting an operation in the API named "operations", we should get the details for that operation.

## Is your portal managed or self-hosted?

Issue present in both.

## Release tag or commit SHA (if using self-hosted version)

Present in managed portal.

## Additional context

This is an issue with function "getResourceName" - The regex it uses will try to get the name for operations, when operations are loaded on initial selection of the API. 

The regex will look for /operations/ and if the API is called operations, the second match will contain a prefix of "operations/..." which is not expected, please see screenshots below:

Regular API name different to "operations":
![image](https://user-images.githubusercontent.com/31003712/220317372-ab9ba9bc-9789-40cf-ab18-2634f9a42fad.png)

API with name set to "operations":
![image](https://user-images.githubusercontent.com/31003712/220317533-2e562a70-7f2b-4bdc-8e83-d5082a6bcaed.png)

Please notice the extra "operations/" in the second screenshot - that is not expected and causing issues.

JMach1:
Hi @cristiteodorov 
thank you for reporting the issue and thank you very much for providing a fix right away!
Unfortunately I'm unable to reproduce the issue. Where exactly it happens? From your repro steps I understood you mean API Detail page, but for me it works just fine with API with name "operations". 
https://jm-apim.developer.azure-api.net/api-details#api=operations

cristiteodorov:
Hi @JMach1 ,

[...]

PR #2113, opened 2023-02-21
Added condition to getResourceName function utility to remove "operations/" from name before returning

cristiteodorov:
This PR includes an edit to the getResourceName function in file Utils.ts to fix the associated issue(#2112).

It allows the developer portal to correctly resolve operations' names for operations in the case of an API named "operations".

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -13,8 +13,11 @@ export class Utils {
         if (matches && matches.length > 1) {
             switch (resultType) {
                 case "name":
+                    // Removes operations/ from name in case of API names "operations"
+                    if (matches[1].includes('operations/')) {
+                        return matches[1].replace('operations/', '');
+                    }
                     return matches[1];
-
                 case "shortId":
                     return `/${resource}/${matches[1]}`;
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -13,7 +13,7 @@ export class Utils {
         if (matches && matches.length > 1) {
             switch (resultType) {
                 case "name":
-                    // Removes operations/ from name in case of API names "operations"
+                    // Removes "operations/"" from name in case of API named "operations"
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -13,8 +13,11 @@ export class Utils {
         if (matches && matches.length > 1) {
             switch (resultType) {
                 case "name":
+                    // Removes "operations/"" from name in case of API named "operations"
+                    if (matches[1].includes('operations/')) {
+                        return matches[1].replace('operations/', '');
+                    }
                     return matches[1];
```
malincrist, line 20: Thanks for the fix! 😄 

I don't think we need the `if` here.
Also, maybe, instead of replacing `operations/` we could replace the resource directly. this way we ensure that the the name is returned correctly for other possible corner cases.
```suggestion
                   case "name":
                      // Removes "<resource>/"" from name in case it matches a reserved word"
                    return matches[1].replace(`${resource}/`, '');
```

JMach1: I guess this might break if the name ends with "resource" string
e.g. /apis/my-api/operations/our-company-operations/foo
the desired result would not be `our-company-foo`

cristiteodorov: Thanks for the suggestions - I can re-test later and come back with a better solution. Or if you have an alternate fix that you want to apply, please let me know and we can cancel this PR.

I think that is right, I did not consider you could have a "/" in the actual operation name as in: "/apis/my-api/operations/our-company-operations/foo" - just to double check, the desired result in this case would be:

"our-company-operations/foo"

cristiteodorov: @JMach1 Just seen your comment on the Issue - closing this.

cristiteodorov:
Fix provided in PR #2122 so closing this.

PR #2132, opened 2023-03-15
Internal customizations for widgets

malincrist:
## Goal
Enable customizations for components of the Developer Portal's widgets.

## Details
Besides the global styles defined for the entire application, users can define local styles for each widget individually.
The components that are stylable are now selectable, and the button edit for "edit local styles" appears when they are selected, as seen in the following screenshot:
image

By clicking on the button, a style editor pop-up will appear, with the allowed customizations for the selected component.
The styles defined will be applied for all of the same components of the widget (i.e. defining a different background color for an input inside a widget will change the background color of all the inputs from inside the same widget)

Here is an example of how styling work for the inputs in the test console:
test-console-tyle


azep-ninja:
When is this expected to go live? This is the last customizations needed to truly give the portal a better overall control of the UI.

PR #2156, opened 2023-04-15
Re-enabled popup controls and snippets.

azaslonov:
In this change we're gettin back the snippet for popup-based Menu and also adding the new "Popups" section to allow managing the popups.


f-alizada:
The PR bumps the version of the paperbits/style. Which results the error of fetching fonts, should it be added somewhere in the documentation how to retrieve/use?

src/libraries/block-snippets.json
```diff
@@ -1581,6 +1594,146 @@
...
+                                                "horizontal": "between"
+                                            }
+                                        }
+                                    },
+                                    "padding": {
+                                        "xs": {
+                                            "top": 10,
+                                            "bottom": 10,
+                                            "left": 10,
+                                            "right": 10
```
JMach1, line 1700: TBH I'm not familiar with this kind of UI definition, but could not be this simplify for something like: padding: { xs: 10 }?

azaslonov: Yes, that would be simpler, but shortcuts not supported by the style compiler yet. Later on, we'll surely simplify it.

azaslonov:
> The PR bumps the version of the paperbits/style. Which results the error of fetching fonts, should it be added somewhere in the documentation how to retrieve/use?

I have added a placeholder for Google Fonts API key and sent the steps to the content team to produce the documentation.

ygrik (approved):

Issue #1461, opened 2021-09-21
Hamburger icon for smaller screen resolutions not working as intended


ZarTrox:
## Bug description

In the DevPortal we are using multiple menu items which are stored in collapsible panels so that when users with smaller screen sizes, e.g. a smartphone, are using this Portal can navigate through the DevPortal with the hamburger Icon.
But when the hamburger icon appears, the Menu is just moved to a different location and does not disappear. See first image.

![image](https://user-images.githubusercontent.com/58420097/134154571-83ee0d86-4010-46ce-b4ff-2821530eed04.png)

Upon clicking on the hamburger icon the menu is correctly displayed. See second image.
![image](https://user-images.githubusercontent.com/58420097/134155841-5b62ab8a-c35a-44fb-a6ed-f5216586c377.png)

Our observation is that the menu which first didn't disappear changes position and is displayed correctly.
We are aware that the new grid approach is in development but since it may take some time until it is finished a solution/workaround for this old problem in the meantime should be present.

## Expected behavior

For smaller screenresolutions the hamburger icon should appear and the menu widget should completely disappear.
When clicking on the hamburger icon the menu widget should be displayed correctly and readable.

## Is your portal managed or self-hosted?

Managed

azaslonov:
This issue has been addressed in Paperbits libraries and will be integrated in next 3-4 weeks along with site heading snippet. This includes full hamburger button customizations, popup container controls, and more.

mrcarlosdev:
This issue is related to managed developer portal. We advise you to create a Azure support request to get assistance on this issue. Please refer to the below link to create a new Azure support request, Please select Problem Type = "Developer Portal" in the request to route it appropriately.

https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

PR #2198, opened 2023-06-04
Upgraded dependencies and default theme.

### LINKED ISSUE #1107: Collapsible menu content alignment bug
While customizing the menu for smaller factors I set the "content alignment".  Now the menu items are always shown regardless of device size.  So now for smaller forms, the menu is shown and the menu button is clickable. When this button is pressed the menu disappears and then reappears. Additionally, while the button is "there" and can be clicked,  it is invisible (possibly related to https://github.com/Azure/api-management-developer-portal/issues/401). 

Is there a way to unset the content alignment? 

The affected portal is hosted by azure.

azaslonov:
Re-designed the website menus to make them popup-based, allowing greater control on how they being styled.
Closes #1107, #1461 (only newly created websites).


ygrik (approved):

templates/default.json
f-alizada, line 1: It seems that the targetKey property is missing from the popup hyperlink configuration which is being processed.

src/services/staticDataHttpClient.ts
```diff
@@ -48,7 +48,12 @@
             response.statusCode = 200;
             response.statusText = "OK";
             response.headers = [{ name: "content-type", value: "application/json; charset=utf-8" }];
-            response.body = Buffer.from(JSON.stringify(this.defaultDataObject[this.getShortUrl(request.url)]));
+            
+            let body = this.defaultDataObject[this.getShortUrl(request.url)];
```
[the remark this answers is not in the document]
f-alizada: fixed

templates/default.json
azaslonov: I see you folks resolved it already.

PR #2208, opened 2023-06-13
Add dropdown for the case when there are multiple examples

f-alizada:
In case there are multiple examples of the request place them under dropdown select.

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/operation-details.html
+++ b/src/components/operations/operation-details/ko/runtime/operation-details.html
@@ -194,11 +194,9 @@
                     <!-- /ko -->
 
                     <!-- ko if: representation.examples.length == 1 -->
-                    <!-- ko foreach: { data: representation.examples, as: 'example' } -->
                     <code-snippet
                         params="{ title: representation.examples[0].title, content: representation.examples[0].value, language: representation.examples[0].format, description: representation.examples[0].description }">
                     </code-snippet>
-                    <!-- /ko -->
                     <!-- /ko -->
 
                     <!-- /ko -->
--- a/src/components/operations/operation-details/ko/runtime/operation-details.html
+++ b/src/components/operations/operation-details/ko/runtime/operation-details.html
@@ -182,7 +182,7 @@
                             </div>
                             <div class="col-4">
                                 <div class="form-group">
-                                    <select class="form-control" data-bind="options: representation.examples, value: $component.selectedRepresentatnionsValue()[representation.contentType], event: {change: $component.permissionChanged} ,optionsValue: 'title', optionsText: 'title', attr : {title: 'Represantation examples selector', id: 'select-' + representation.contentType + '_' + $index() }"></select>
+                                    <select class="form-control" data-bind="options: representation.examples, value: $component.selectedRepresentatnionsValue()[representation.contentType] ,optionsValue: 'title', optionsText: 'title', attr : {title: 'Representation examples selector', id: 'select-' + representation.contentType + '_' + $index() }"></select>
                                 </div>
                             </div>
                         </div>
@@ -196,7 +196,7 @@
                     <!-- ko if: representation.examples.length == 1 -->
                     <!-- ko foreach: { data: representation.examples, as: 'example' } -->
                     <code-snippet
-                        params="{ title: example.title, content: example.value, language: example.format, description: example.description }">
+                        params="{ title: representation.examples[0].title, content: representation.examples[0].value, language: representation.examples[0].format, description: representation.examples[0].description }">
                     </code-snippet>
                     <!-- /ko -->
                     <!-- /ko -->
--- a/src/components/operations/operation-details/ko/runtime/operation-details.html
+++ b/src/components/operations/operation-details/ko/runtime/operation-details.html
@@ -175,11 +175,32 @@
                     <!-- /ko -->
 
                     <!-- ko if: representation.examples -->
+                    <!-- ko if: representation.examples.length > 1 -->
+                        <div class="row flex flex-row example-form-control">
+                            <div class="col-12">
+                                <label data-bind="attr : { for: 'select' + representation.contentType + '_' + $index() }, text: 'Examples'"></label>
+                            </div>
+                            <div class="col-4">
+                                <div class="form-group">
+                                    <select class="form-control" data-bind="options: representation.examples, value: $component.selectedRepresentatnionsValue()[representation.contentType], event: {change: $component.permissionChanged} ,optionsValue: 'title', optionsText: 'title', attr : {title: 'Represantation examples selector', id: 'select-' + representation.contentType + '_' + $index() }"></select>
+                                </div>
+                            </div>
+                        </div>
+                        <!-- ko foreach: { data: representation.examples, as: 'example' } -->
+                        <code-snippet data-bind="visible: $component.selectedRepresentatnionsValue()[representation.contentType]() == example.title"
+                            params="{ title: example.title, content: example.value, language: example.format, description: example.description }">
+                        </code-snippet>
+                        <!-- /ko -->
+                    <!-- /ko -->
+
+                    <!-- ko if: representation.examples.length == 1 -->
                     <!-- ko foreach: { data: representation.examples, as: 'example' } -->
                     <code-snippet
                         params="{ title: example.title, content: example.value, language: example.format, description: example.description }">
                     </code-snippet>
                     <!-- /ko -->
+                    <!-- /ko -->
+
                     <!-- /ko -->
                 </div>
             </div>
--- a/src/components/operations/operation-details/ko/runtime/operation-details.ts
+++ b/src/components/operations/operation-details/ko/runtime/operation-details.ts
@@ -237,12 +237,9 @@ export class OperationDetails {
 
         if (operation) {
             await this.loadDefinitions(operation);
-            if (this.showExamples) this.parseExamples(operation);
+            if (this.showExamples) this.parseResponseExamples(operation);
 
-
-            if(!operation) return null;
-
-            this.loadExamples(operation);
+            this.loadRequestExamples(operation);
 
             this.operation(operation);
         }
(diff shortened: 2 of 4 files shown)


src/components/operations/operation-details/ko/runtime/operation-details.html
```diff
@@ -175,12 +175,33 @@ <h4>Request body</h4>
...
 
                     <!-- ko if: representation.examples -->
+                    <!-- ko if: representation.examples.length > 1 -->
+                        <div class="row flex flex-row example-form-control">
+                            <div class="col-12">
+                                <label data-bind="attr : { for: 'select' + representation.contentType + '_' + $index() }, text: 'Examples'"></label>
+                            </div>
+                            <div class="col-4">
+                                <div class="form-group">
+                                    <select class="form-control" data-bind="options: representation.examples, value: $component.selectedRepresentatnionsValue()[representation.contentType], event: {change: $component.permissionChanged} ,optionsValue: 'title', optionsText: 'title', attr : {title: 'Represantation examples selector', id: 'select-' + representation.contentType + '_' + $index() }"></select>
```
DeagleGross, line 185: typo `Represantation`

src/components/operations/operation-details/ko/runtime/operation-details.html
```diff
@@ -175,12 +175,33 @@ <h4>Request body</h4>
...
+                                </div>
+                            </div>
+                        </div>
+                        <!-- ko foreach: { data: representation.examples, as: 'example' } -->
```
DeagleGross, line 189: what is the purpose of multiple "ko ..." existing in the file?

DeagleGross: as comments

Mielek: This is Knockout framework syntax to create repeated html elements. for example create lists of elements which are dynamically fetched from server

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -232,6 +238,12 @@ export class OperationDetails {
         if (operation) {
             await this.loadDefinitions(operation);
             if (this.showExamples) this.parseExamples(operation);
+
+
+            if(!operation) return null;
+
+            this.loadExamples(operation);
```
Mielek, line 245: Why load examples is after parseExamples?

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -232,6 +238,12 @@ export class OperationDetails {
         if (operation) {
             await this.loadDefinitions(operation);
             if (this.showExamples) this.parseExamples(operation);
+
+
+            if(!operation) return null;
```
Mielek, line 243: If operation is not changed in the functions then it will always not be falsy. See line 238.

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -244,6 +256,19 @@ export class OperationDetails {
...
+        const representations = operation.request.meaningfulRepresentations().map(a => {return {...a}});
+        let requestExamples = {};
+        if (representations && representations.length) {
+            for(var i = 0; i < representations.length; i++) {
+                let value = representations[i].examples?.[0];
+                if (!value) return;
+                requestExamples[representations[i].contentType] =  ko.observable(value.title);
+            }
+        }
+        this.selectedRepresentatnionsValue(requestExamples);
```
Mielek, line 269:
```suggestion
        const requestExamples = operation.request.meaningfulRepresentations()
        .map(a => {return {...a}})
        .filter(a => a.examples?.[0])
        .reduce( (a, c) => {
          c[a.contentType] = ko.observable(a.title);
          return c;
        }, {});
        this.selectedRepresentatnionsValue(requestExamples);
```

Mielek (changes requested):
Please invoke lint fix / prettier fix for styling

src/components/operations/operation-details/ko/runtime/operation-details.ts
Mielek: This is only a suggestion to use build in functions to accomplish same result  :) No need to apply.

src/components/operations/operation-details/ko/runtime/operation-details.html
```diff
@@ -175,12 +175,33 @@ <h4>Request body</h4>
...
+                        <code-snippet data-bind="visible: $component.selectedRepresentatnionsValue()[representation.contentType]() == example.title"
+                            params="{ title: example.title, content: example.value, language: example.format, description: example.description }">
+                        </code-snippet>
+                        <!-- /ko -->
+                    <!-- /ko -->
+
+                    <!-- ko if: representation.examples.length == 1 -->
```
malincrist, line 196: since `representation.examples.length == 1`, why do we foreach `representation.examples`? I think it would make more sense to just use `representation.examples[0]`

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -244,6 +256,19 @@ export class OperationDetails {
         this.working(false);
     }
 
+    public async loadExamples(operation: Operation): Promise<void> {
+        const representations = operation.request.meaningfulRepresentations().map(a => {return {...a}});
```
malincrist, line 260: why do we need this copy of the object? can't we use directly `operation.request.meaningfulRepresentations()` ?

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -244,6 +256,19 @@ export class OperationDetails {
...
+        let requestExamples = {};
+        if (representations && representations.length) {
```
malincrist, line 262: instead of if and for, you could use something like
`representations?.foreach((representation, index) => {...}`

If you want to go with the for approach, I think it would be better to use `let` instead of `var` here. 
The variables that are defined with `let` statement have block scope.
The variables that are defined with `var` statement have function scope.
some reference: https://levelup.gitconnected.com/const-vs-let-vs-var-in-javascript-which-one-should-you-use-c56cf9b9e2a3

src/components/operations/operation-details/ko/runtime/operation-details.html
f-alizada: this is knockout js code, which executes as part of the page. when you need some logic there to hide something to loop the array

src/components/operations/operation-details/ko/runtime/operation-details.html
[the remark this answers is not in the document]
f-alizada: nothing removing, was testing something :) Thank you

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: Thank you for the suggestion. 
May I clarify if this is mandatory change? 
In terms of the performance is there any difference?

Mielek: I is not a mandatory change.

I do not think there is a major difference in performance

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -244,6 +256,19 @@ export class OperationDetails {
...
+            for(var i = 0; i < representations.length; i++) {
+                let value = representations[i].examples?.[0];
+                if (!value) return;
```
Mielek, line 265: Why do you finish the function here? should you not just skip the iteration?

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: parse examples actually also does operations over the examples from the response. Let me rename it thanks

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: Ideally it should not be the case, but let me rewrite it, thanks

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: true, thanks. left things after testing :)

src/components/operations/operation-details/ko/runtime/operation-details.html
f-alizada: Thank you updated

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: updated thank you

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: updated thank you

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: agree thank you.

Mielek (approved):

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -346,11 +368,11 @@ export class OperationDetails {
         return result;
     }
 
-    private parseExamples(operation: Operation): void {
+    private parseResponseExamples(operation: Operation): void {
         const examples = operation.getMeaningfulResponses().reduce((acc, cur) => {
             const representations = cur.meaningfulRepresentations();
             if (!representations || !representations.length) return acc;
-
+
```
Mielek, line 375: It is minor. Style.

Mielek (approved):

src/components/operations/operation-details/ko/runtime/operation-details.ts
f-alizada: Let me merge that in such case and will be fixing in upcoming PR. on top of it, thank you!

malincrist (approved):

Issue #2202, opened 2023-06-06
Incorrect message for invalid/missing subscription key using policy fragment with cors & Users shouldn't be able to send requests when not subscribed.


Ledger94:
**Every bug report should have precise description and reproduction steps; console traces or source code references are appreciated.**

**For assistance requests, contact Azure support or submit a post [on Stack Overflow](http://aka.ms/apimso). We don't provide support through GitHub Issues. Feature requests can be raised [on the Azure Feedback Forum](https://aka.ms/apimwish).**

## Bug description

1. Incorrect message is showing when the subscription key is incorrect or missing. Cors policy has been set using policy fragment and it works as expected with the correct subscription key. **This isn't an issue when not using policy fragments.** (setting Cors in API scope without policy fragment)

Issue:
![image](https://github.com/Azure/api-management-developer-portal/assets/84686660/d2a4f75e-1e4d-4fe8-9600-1ea91d573fc1)

Expected Response:
![image](https://github.com/Azure/api-management-developer-portal/assets/84686660/fa92e487-143a-4bfc-9cb5-b22a8e63a7d4)

2. Users shouldn't be able to send requests for APIs within products that require subscription keys, it is possible to do this currently. There should be some kind of message/error letting the user know that a subscription is required for the API like in old portal.

New Portal Ex:
![image](https://github.com/Azure/api-management-developer-portal/assets/84686660/6f1d84c1-7852-40c2-8ed5-4947ad6fdf3c)

Old Portal Ex:
![image](https://github.com/Azure/api-management-developer-portal/assets/84686660/997c9859-55e9-4cd4-8555-50297b530b71)

## Reproduction steps

First Bug:
1. Create policy fragment that has cors policy.
2. Add cors policy fragment to API scope.
3. Go to api-details page for an API that you are subscribed to test with correct subscription key. (It should work).
4. Test with incorrect subscription key (Cors error).
5. Update Cors without policy fragement.
6. Test again with incorrect subscription key (expected behavior 401 response).

Second Bug:
1. Go to api-details page for an API whos product requires a subscription key and you are not subscribed to it.
2. Click try it for pop up.
3. Test API without inputting subscription key (It should allow you to send request).

## Expected behavior

1. Incorrect message is displaying related to cors when using a policy fragment that contains cors policy. It should work the same whether or not cors is in a policy fragment.
2. For APIs whos product requires a subscription key, it shouldn't be possible to test it without inputting a subscription key. 

[...]

PR #2212, opened 2023-06-15
Subscription key improvements

malincrist:
This PR introduces some improvements for the Subscription Key flows in the test console:
1. If the subscription key is required, the header is always present in the "headers" section, and marked as required:
image

2. The "Subscription key" from Authorization section is masked out by default an it can be un-masked:
subscriptionkeyauth

3. If the user tries to send the request without any value completed for the subscription key header, the request will not be sent and an error will be shown: (related issue #2202)
image

Similar behavior for GraphQL and WebSocket and subscription key parameter.

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/authorization.html
+++ b/src/components/operations/operation-details/ko/runtime/authorization.html
@@ -11,12 +11,12 @@
     <div class="row flex flex-row">
         <div class="col-4">
             <label for="authServer" class="text-monospace form-label"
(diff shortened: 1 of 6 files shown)


f-alizada (approved):

arbs-io:
Hi @f-alizada, @malincrist 
Has this change introduced a bug? We've noticed over the weekend that the subscription key is now always required. Even if both Product and API have subscription-required disabled.

PR #2214, opened 2023-06-20
API details page

malincrist:
This PR implements a new widget called the "API details page" which is designed to enhance and replace the existing layout for the APIS page.

The widget features a left-side menu that allows users to access various API details, including products, changelog, and operations. The customizations for the operations within this widget are consistent with those found in the "List of Operations" widget, while the customizations for the operation details match those of the "Operation details" widget.

Additionally, this update introduces a new section called "Documentations" which showcases the markdown documentations associated with the selected API.

Below is a screenshot of the widget:
image


ronaldbosma:
I'm interested in shownig documentations on the API details page and perhaps on other pages as well. Is there any information on when this change will be available in the developer portal?

PR #2254, opened 2023-07-24
Playwright tests

f-alizada:


tests/mocks/collection/user.ts
```diff
@@ -1,108 +1,74 @@
...
     
-    public getSignInResponse(){
+    public constructor(testId: string, email: string, publicId: string, firstName: string, lastName: string, password: string){
+        super(testId);
+        this.email = email;
+        this.publicId = publicId;
+        this.firstName = firstName;
+        this.lastName = lastName;
+        this.password = password;
+        this.accessToken = Utils.getSharedAccessToken(this.publicId, "accesskey", 1);
```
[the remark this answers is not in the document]
f-alizada: removed

tests/services/IApiService.ts
```diff
@@ -0,0 +1,9 @@
+import { ApiContract } from "../../src/contracts/api";
+export interface IApiService {
```
ygrik, line 2: May be better to named ITestApiService to avoid mixing with not testing services?

tests/utils.ts
```diff
@@ -1,10 +1,9 @@
 import * as fs from "fs";
 import * as crypto from "crypto";
 import * as http from "http";
-import { ConsoleMessage, Page } from 'puppeteer';
 
 export class Utils {
```
ygrik, line 5: it's easy to misuse testing Utils

tests/services/IApiService.ts
f-alizada: thank you for review

tests/utils.ts
f-alizada: updated all the services and utils names to prepend the test Prefix, thank you for the review

src/services/usersService.ts
```diff
@@ -255,7 +255,7 @@ export class UsersService {
      */
     public async ensureSignedIn(): Promise<string> {
         const userId = await this.getCurrentUserId();
-
+
```
Mielek, line 258: Leftover whitespace

tests/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
```
Mielek, line 1: you can use the `extend` functionality of tsconfig to not duplicate properties here. See https://www.typescriptlang.org/tsconfig#extends

playwright.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  testIgnore: 'playwright/*',
+  //timeout: 50_000,
```
Mielek, line 5: Leftover?

tests/e2e/runtime/products.spec.ts
```diff
@@ -1,42 +1,35 @@
...
-        await browser.close();
-        Utils.closeServer(server);
-    });
-
-    it("User can see producst on the page", async () => {
-        var products = new Products();
-        products.addProduct(Product.getStartedProduct());
-        products.addProduct(Product.getUnlimitedProduct());
+test.describe("products-page", async () => {
+    test("published-products-visible-to-guests", async function ({page, configuration, cleanUp, mockedData, productService, testRunner})  {
```
Mielek, line 7: I think what we can do to ease the devs in creating the test we can provide them with small helper to create all of mock data preparation, data population, and validation. Please consider that.

Mielek (approved):

tests/tsconfig.json
f-alizada: Yes it is an options, however I would like to keep test related tsconfig separately and writtent explicitly, in case there will be some change in the main file it should not affect all tests.

src/services/usersService.ts
f-alizada: removed thanks

playwright.config.ts
f-alizada: removed thanks

tests/e2e/runtime/products.spec.ts
f-alizada: I would like to keep the changes for the next iteration, reasons behind that: 
1 - To see the usage of the test by developers and gather more feedback
2 - To understand main bottlenecks while implementing the test

Mielek (approved):

PR #2274, opened 2023-08-16
Support downloading excel and csv responses with content disposition attachment header

LauraXiulan:
When retrieving an CSV or EXCEL file as response in the Developer Portal, the response is shown inline, which makes it unreadable. This PR adds logic to download these kinds of responses instead.

Suggestions, remarks and edits are welcome. I tested the code locally for our own APIM by temporary switching to Self-Hosted.
I have only added CSV and EXCEL in this PR, but I can imagine a usecase for PDF's as well.

### DIFF
--- a/src/components/operations/operation-details/ko/runtime/operation-console.ts
+++ b/src/components/operations/operation-details/ko/runtime/operation-console.ts
@@ -1,7 +1,7 @@
 import * as ko from "knockout";
 import * as validation from "knockout.validation";
 import { ISettingsProvider } from "@paperbits/common/configuration";
-import { HttpClient, HttpMethod, HttpRequest } from "@paperbits/common/http";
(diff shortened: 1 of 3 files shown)


optyingale (approved):

optyingale (approved):

LauraXiulan:
@ygrik @malincrist @azaslonov

verschaevesiebe:
Nice addition, we sometimes have the intent to download other content types. Eg. a very large JSON file for example.
Would you be able to include a similar feature for attachment disposition types where filenames are json or by means any text formal is applicable ?

E.g xml, text, json, jsonc, etc ...

header('Content-Disposition: attachment; filename=data.json');
header('Content-Type: application/json');

Would be amazing thanks !

PR #2313, opened 2023-10-13
Added validations for passwords

malincrist:
Added validations for the passwords on the front-end that match the validations from the back-end.
password


azaslonov:
Closing a stale.

PR #2330, opened 2023-11-06
feat: add option to configure a default to expand all group tags

### LINKED ISSUE #2119: [Feature request] Show tag operations by default
Tag operations are now hidden by default after the previous release (https://github.com/Azure/api-management-developer-portal/pull/2003).

This is useful when having a long list of tags but it looks confusing if an API have only one or two tags.

Is it possible to add an option to the "List of operations" widget to change the default behaviour to expand all tag sections by default?

Before:
Screenshot 2023-02-27 at 14 58 17

Now:
Screenshot 2023-02-27 at 14 58 26

jan-mrm:
Adds a designer option in order to configure a default to expand the group tags in the api view.

somewhat closes #2119 since you now at least can configure it and go back to having all groups expanded

I did let the initial default to be `not expanded` (false). Not sure whats your stance on this is, or if it should / could also be changed to `expanded` (true)

### DIFF
--- a/src/components/operations/operation-list/ko/operationListEditor.html
+++ b/src/components/operations/operation-list/ko/operationListEditor.html
@@ -34,6 +34,13 @@
     </div>
 
     <div class="form-group">
+        <label for="defaultAllGroupTagsExpanded" class="form-label">
+            <input type="checkbox" id="defaultAllGroupTagsExpanded" name="defaultAllGroupTagsExpanded" data-bind="checked: defaultAllGroupTagsExpanded" />
+            Default to expand all GroupTags
+        </label>
+    </div>
+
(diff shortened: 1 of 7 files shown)


malincrist:
@jan-mrm , thank you for your contribution! 😀

malincrist (approved):

jan-mrm:
sure thing 🙂
can you point me to the root of the error from the failed check? Or is it not something that is a result of this pull request?

malincrist:
The accessibility scan failed because this PR is made from a forked repository, and there is an issue with the scan in such scenarios. Your changes don't seem related.

jan-mrm:
alright, thanks 🙂

PR #2337, opened 2023-11-23
Added UI interface for multiple auth servers

malincrist:
This PR introduces displaying the authorization servers specified for an API:

The servers displayed are only the ones that have the property "useInApiDocumentation" set to true.

The properties displayed for the servers in the operation details are the following:
- Name (only if multiple servers are specified) 
- Description 
- Supported grant flows 
- Authorization endpoint URL 
- Token endpoint URL 
- Scopes 

Here is an example of how multiple authorization servers are displayed:
aaaalign

The servers used for the test console are the ones that have the property "useInTestConsole" set to true:
In the test console, there is a new selector, where the authorization server can be selected. 
The authentication flow remains the same: when the selected grant type is changed, the authentication flow is initiated.
multiple-auth-test-console


src/components/operations/operation-details/ko/runtime/authorization.html
```diff
@@ -7,17 +7,38 @@ <h3 class="pt-0">
         </h3>
     </summary>
 
-    <!-- ko if: $component.authorizationServer -->
+    <!-- ko if: $component.authorizationServers()?.length > 0 -->
     <div class="row flex flex-row">
         <div class="col-4">
-            <label for="authServer" class="text-monospace form-label"
-                data-bind="text: $component.authorizationServer().displayName"></label>
+            <label for="authSerer" class="text-monospace form-label">Authorization server: </label>
```
jsorohova, line 13: A typo in "for" attribute.

malincrist: fixed, thanks! :)

jsorohova (approved):

PR #2368, opened 2024-01-17
Fixed error when changing OAuth flow

malincrist:
### Problem
In the test console, after a successful authentication with "authorization_code" flow, if the user tries to authenticate with "authorization_Code (PKCE)", the error "Unable to authenticate due to internal error." is thrown.
image

### Solution
Remove the event listener after the authentication with either of the flows is completed, so they don't overlap.

### DIFF
--- a/src/services/oauthService.ts
+++ b/src/services/oauthService.ts
@@ -120,41 +120,24 @@ export class OAuthService {
             query: query
         });
 
-        return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


rkolesnikovDX:
One more question: we can code path when something go wrong during auth. Don't we need add code that remove listener there too?

rkolesnikovDX (approved):

PR #2375, opened 2024-01-23
Fixed: error when loading subscriptions without user signed in

malincrist:
### Problem
When initializing the "Subscriptions" widget, we check if the user is signed in and if not, we redirect them to the sign in page. However, the flow for loading subscriptions continues, even if there is no userId set, resulting in the error `Unable to load subscriptions. Error: Parameter "userId" not specified.`

### Solution
If, after checking whether the user is signed in or not, we do not receive any userId in return, stop the execution of the initialization of the widget.

### DIFF
--- a/src/components/users/subscriptions/ko/runtime/subscriptions.ts
+++ b/src/components/users/subscriptions/ko/runtime/subscriptions.ts
@@ -54,6 +54,10 @@ export class Subscriptions {
 
     private async loadUser(): Promise<void> {
         this.userId = await this.usersService.ensureSignedIn();
+        if(!this.userId){
+            return;
+        }
+        
         await this.loadSubscriptions();
     }
 


rkolesnikovDX (commented):
We have similar code in https://github.com/Azure/api-management-developer-portal/blob/3e2d250139905f68799d39705344a07c5b38b0fc/src/components/users/profile/ko/runtime/profile.ts#L67

Don't we need to have a similar check and throw readable error there?

malincrist:
> We have similar code in
> 
> https://github.com/Azure/api-management-developer-portal/blob/3e2d250139905f68799d39705344a07c5b38b0fc/src/components/users/profile/ko/runtime/profile.ts#L67
> 
> Don't we need to have a similar check and throw readable error there?

I don't think so, in the subscriptions widget, the issue is that we set ` this.userId = await this.usersService.ensureSignedIn();` and then rely on the userId to load the subscriptions.

In the profile widget we do this check just to redirect to sign in incase the user is not signed in, but we don't actually use the userId later **for the initialization of the widget**. 
We use the userId only for specific actions triggered by the user. However the data for the user is taken from the next line of code `const model: User = await this.usersService.getCurrentUser();`, and this is based on the stored token. So, if the user is not signed in, there is no token, hence the user model is null. And since we have `<!-- ko if: user -->` in the html, we are safe because the buttons are not even displayed, so the user cannot trigger any action (super fast) before the redirect actually happens.

I don't think it's necessary to add it in the profile widget as well, except for maybe keeping consistency in the code. What do you think?

rkolesnikovDX (approved):

Issue #2326, opened 2023-10-30
Cell text truncated because there is no wrap when using "oneof"


josealbertocoronado:
## Bug description
When using "oneof" in yaml for an object, if the list of elements is long it gets truncated because the cell style has no wrapping. 

## Reproduction steps
1. Go to resource https://ig-weu-dev-integration-apim.developer.azure-api.net/
2.Go to "Inbound-Outbound API - v2.0"
3. Scroll down to `multiTypePayload` object
4. See "OneOf" cell is truncated so not all elements show up:
![image](https://github.com/Azure/api-management-developer-portal/assets/132665924/97339adb-3f03-4df1-ba67-cb0d679482f7)

## Expected behavior
The content is wrapped so it can be visualized entirely (maybe adding ".text-truncate { white-space: unset !important; }" ?)

## Is your portal managed or self-hosted?
Managed

## API Management service name
ig-weu-dev-integration-apim

## Environment
 - Operating system:Win 11
 - Browser: Both Edge and Chrome

malincrist:
Hello @josealbertocoronado , thanks for reporting this issue. We have created an internal work item to address this and we will keep you updated on the progress.

josealbertocoronado:
Lovely. Thanks for that.

josealbertocoronado:
Hi @malincrist.
Any ETA for this? Even when it looks a minor issue is a bit blocking, as users can click the links and therefore they can't easily navigate to the desired object.

josealbertocoronado:
Hi @malincrist .
Sorry chasing, is there any plan to release a fix for this issue soon?
Cheers.

josealbertocoronado:
Hi @jsorohova .
I see this ticket has been completed now but if I check my resource I still can see the old behaviour.
Is it maybe not deployed yet?
Cheers.

Issue #2290, opened 2023-09-14
Links to some elements don't work for objects with inheritance


josealbertocoronado:
**Every bug report should have precise description and reproduction steps; console traces or source code references are appreciated.**

**For assistance requests, contact Azure support or submit a post [on Stack Overflow](http://aka.ms/apimso). We don't provide support through GitHub Issues. Feature requests can be raised [on the Azure Feedback Forum](https://aka.ms/apimwish).**

## Bug description

For some properties in an object that inherits from a different object, the links to some of the properties don't work. The element is not even loaded in the page.

## Reproduction steps

1. Go to ig-weu-dev-integration-apim developer portal.
2. Go to the API called "Gateway API Services".
3. Go to the "multitypeRequest" endpoint.
4. Navigate to "firstNotificationMotor" object.
5. Click in "brokerReference" property, link doesn't work. Interestingly some other similar fields work smoothly (f.i. "brokerName").

## Expected behavior

Links to the different elements in a object always work.

## Is your portal managed or self-hosted?

Managed

## API Management service name

ig-weu-dev-integration-apim

JMach1:
Hello @josealbertocoronado 
I reproduced the issue, we'll work on fixing the bug. 
ETA next portal release in ~ 1 month 
Thank you for reporting it!

josealbertocoronado:
Lovely. Thanks for that.

josealbertocoronado:
Hi, sorry revisiting this ticket but unfortunately the fix hasn't worked.

Links for properties like "brokerReference" (see description above) still don't work.

Can you please take a look? Cheers.

josealbertocoronado:
Hi @jsorohova , any update on this, please? It's quite blocking now, as users complain about the bad experience they're getting with the developer portal when they can't navigate to the different properties.

josealbertocoronado:
Hi @jsorohova , I'm not sure whether it's the same issue or something new but now I'm experiencing a different issue with links.
No single link works now. Whatever link I click it always takes me to the top of the page. It's happening since not too long ago, so i reckon it's related to the last release.

You just reproduce it by going to our company portal and clicking any link: https://ig-weu-dev-integration-apim.developer.azure-api.net/

This is quite blocking as the image we expose to our customers is quite poor, so I would really appreciate if you can take a look with some urgency, please.

Cheers.

josealbertocoronado:
Morning @jsorohova .
[...]

PR #2389, opened 2024-01-31
Fixed non-working anchors for definitions, made multiple complex schemas more readable

jsorohova:
Closes: #2326, #2290

Problems:
1. Definition anchors were not working, all links were redirecting to the top of the page.
2. When multiple definitions were listed in complex schemas, they were truncated.

Solutions:
1. Fixed scrolling so it scrolls to the definition.
2. Listed every definition on the new line so they are not truncated.


src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -454,8 +454,14 @@ export class OperationDetails {
     }
 
     private scrollToOperation() {
```
malincrist, line 456: I think we should rename this method, since it's not used only for scrolling to the operation, but rather any scrolling inside the operation details. What do you think?

src/components/operations/operation-details/ko/runtime/operation-details.ts
```diff
@@ -454,8 +454,14 @@ export class OperationDetails {
...
-        const headerElement = document.getElementById("operation-name");
-        headerElement && headerElement.scrollIntoView({ behavior: "smooth", block: "start", inline: "start" });
+        if (this.routeHelper.getDefinitionName()) {
+            const definitionId = this.router.getHash();
```
malincrist, line 458: Here, I think the code would be cleaner if we use the if statement just for deciding where the scroll should happen, and then do the actual scrolling. Something like:


```
let scrollTo : HTMLElement;

if (this.routeHelper.getDefinitionName()) {
    const definitionId = this.router.getHash();
    scrollTo = document.getElementById(definitionId);
} else {
    scrollTo = document.getElementById("operation-name");
}

scrollTo && scrollTo.scrollIntoView({ behavior: "smooth", block: "start", inline: "start" });
    
```

malincrist (approved):

malincrist (approved):

PR #2402, opened 2024-02-12
Improved logging around AAD B2C flows

malincrist:
This PR enhances the logs around AAD B2C flows:
- Added more logs around the actual flow (i.e. the flow was initialized, successful login etc.)
- Log errors from `parseAndDispatchError`: our current AAD B2C errors are only shown to users from this function. Added tracker logs here so we know from logs what happened.
- Added `traceClick` binding handler, so we can trace user clicks.

### DIFF
--- a/src/components/users/signin-social/ko/runtime/signin-aad-b2c.ts
+++ b/src/components/users/signin-social/ko/runtime/signin-aad-b2c.ts
@@ -8,6 +8,8 @@ import { AadB2CClientConfig } from "../../../../../contracts/aadB2CClientConfig"
 import { AadService, AadServiceV2, IAadService } from "../../../../../services";
 import { dispatchErrors, parseAndDispatchError } from "../../../validation-summary/utils";
 import { ErrorSources } from "../../../validation-summary/constants";
+import { Logger } from "@paperbits/common/logging";
+import { eventTypes } from "../../../../../logging/clientLogger";
 
 
 const aadb2cResetPasswordErrorCode = "AADB2C90118";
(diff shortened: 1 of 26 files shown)


src/components/users/signin-social/ko/runtime/signin-aad-b2c.ts
```diff
@@ -60,15 +63,22 @@ export class SignInAadB2C {
             }
             await this.selectedService.checkCallbacks();
         }
+        else {
+            this.logger.trackEvent(eventTypes.aadB2CLogin, { message: "AAD B2C client configuration is missing." });
+        }
+
+        this.logger.trackEvent(eventTypes.aadB2CLogin, { message: "Signin AAD B2C component initialized." });
```
f-alizada, line 70: Shouldn't this event be tracked inside the
`if (this.aadConfig) {` when the config actually presents?  Otherwise in case when the configs are not presenting which is an error scenario, but the logs will report `component initialized`.

malincrist: I'm not sure.. Ideally, we shouldn't kick off the initialization of this widget if the config isn't present. However, in this situation, even if we somehow initiated the process without a config, it would still go through. It's an unusual scenario, but as long as it doesn't trigger an error, the initialization can wrap up fine.
In this case, I think it's important to have both logs, so we know that the initialization was completed without the config.

I wouldn't treat t as an error scenario, because there was no error thrown. Maybe we could treat it as a warning since indeed the widget doesn't make any sense without the config (though, at the moment, we don't have support for different log levels from the client side). What are your thoughts on this?

f-alizada: If we are referring to the component overall yes I do agree that initialized at the end is a good information! :) Thank you

f-alizada (approved):

PR #2424, opened 2024-02-28
Fixed: API not loaded the first time

### LINKED ISSUE #2421: "the specific api does not exist" shown if API is selected in `api-details` widget
## Bug description

The developer portal widget `api-details` shows "the specific api does not exist" if you initially select an API in the `api-list (Dropdown)` widget. Even though a valid API has been selected and the API exists.

If you refresh the page after selecting the API, the `api-details`  shows the correct information.

## Reproduction steps

1. In the developer portal management ui create a new page containing the `api-list (Dropdown)` widget and a `api-details` widget.
  image
2. Save and publish
3. Open the just created page in the developer portal as a "normal" user (not the management ui)
4. In API Details component there is a text "The specified API does not exist." even though no API was selected.
5. Select an API in the dropdown
6. the `api-details` shows the message "the specific api does not exist" instead of the actual API details.
  image
8. if you refresh the browser window the API details are shown correctly
  image

Once the browser window has been refreshed the dropdown works as expected and changes the `api-details` content upon change.

## Expected behavior

- Before selecting an API from the list, different information than "The specified API does not exist." is displayed because in current approach it is kind of misleading that specified API does not exist even though no API is selected.
- Title in API Details component is populated right away, without refresh required after selecting an API.

Managed

## Environment

 - Operating system: MacOS
 - Browser: Firefox 
 - Version: 122.0.1

## Additional context

none

malincrist:
### Problem
When navigating on a page that has `API: Details` widget, the first time the user selects an API (from the widget API: list for example), it is not loaded.
This is happening due to the order of the operations in the initialization flow: if there is no API in the URL initially we `return` without subscribing to `onRouteChange()`. Hence, even if the route changes, the API: Details widget doesn't react.

### Solution
Subscribe to `onRouteChange()` even if there is no API in the URL.

Closes #2421

### DIFF
--- a/src/components/apis/details-of-api/ko/runtime/api-details.ts
+++ b/src/components/apis/details-of-api/ko/runtime/api-details.ts
@@ -44,7 +44,8 @@ export class ApiDetails {
     @OnMounted()
     public async initialize(): Promise<void> {
         const apiName = this.routeHelper.getApiName();
-
+        this.router.addRouteChangeListener(this.onRouteChange);
+        
         if (!apiName) {
             return;
         }
@@ -52,7 +53,6 @@ export class ApiDetails {
         this.selectedApiName(apiName);
         await this.loadApi(apiName);
 
-        this.router.addRouteChangeListener(this.onRouteChange);
         this.currentApiVersion.subscribe(this.onVersionChange);
         this.downloadSelected.subscribe(this.onDownloadChange);
     }


MarvinVaillant:
Thanks a lot for the fix :)

mpodriezov (approved):

PR #2436, opened 2024-03-15
Update issue templates

sthirthala:


annaji-msft (commented):
please remove this option completely from the template 
image

annaji-msft (approved):

PR #2495, opened 2024-05-28
Reverting script to automate reverting the editor from a published developer portal revision

mpodriezov:
Adding a simple script which is going to help to revert a snapshot from the tenant's storage back to editing.


azaslonov:
Closing a stale.

PR #2500, opened 2024-06-04
FUI - Redesign toggle

jsorohova:
image


src/admin/leftPanel.tsx
```diff
@@ -217,6 +233,18 @@ export class LeftPanel extends React.Component<{}, LeftPanelState> {
                     { this.state.selectedNavItem === NavItem.Media && <MediaModal onDismiss={this.handleBackButtonClick.bind(this)} /> }
                     { this.state.selectedNavItem === NavItem.Settings && <SettingsModal onDismiss={this.handleBackButtonClick.bind(this)} /> }
                     { this.state.selectedNavItem === NavItem.Help && <HelpModal onDismiss={this.handleBackButtonClick.bind(this)} /> }
+                    <Toggle
+                        label={"Enable runtime redesign"}
```
JMach1, line 237: I think we'll have to think about a better name, let's talk with Sreekanth on standup

PR #2509, opened 2024-06-13
Update migrate.yml

TaoChan:
Fixed variable names that are wrong and prevented running the pipeline successfully


TaoChan:
Currently the pipeline is broken due to variable references are wrong

PR #2514, opened 2024-06-19
browser specific iframe sandbox tokens support

JMach1:
Remove iframe sandbox tokens which are not supported in Safari to separated string and add based on if the browser supports them. This way we won't break existing customer widgets.

### DIFF
--- a/src/components/custom-widget/ko/runtime/customWidget.ts
+++ b/src/components/custom-widget/ko/runtime/customWidget.ts
@@ -23,7 +23,7 @@ export class CustomWidget {
         this.instanceId = ko.observable();
 
         const iframe = document.getElementsByTagName("iframe")[0];
-        this.iframeSandboxAllows += " " + iframeSandboxAllowsBrowserSpecific
+        this.iframeSandboxAllows = (iframeSandboxAllows + " " + iframeSandboxAllowsBrowserSpecific)
             .split(" ")
             .filter(token=> iframe?.sandbox.supports(token))
             .join(" ");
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -316,7 +316,8 @@ export const iframeAllows = "clipboard-read; clipboard-write; camera; microphone
 /**
  * List of allowed attributes for a sandboxed iframe.
  */
-export const iframeSandboxAllows = "allow-same-origin allow-scripts allow-modals allow-forms allow-popups allow-popups-to-escape-sandbox allow-top-navigation allow-pointer-lock";
+export const iframeSandboxAllows = "allow-scripts allow-modals allow-forms allow-popups allow-popups-to-escape-sandbox allow-top-navigation allow-pointer-lock";
+export const iframeSandboxAllowsBrowserSpecific = "allow-downloads allow-presentation allow-orientation-lock";
```
ygrik, line 320: why do you use two const if you always check support by the browser for each item?

JMach1: iframeSandboxAllows is still used in for example Custom Widgets editor, where I really don't expect they would need "allow-downloads" and it would be waste of performance making it runtime for this

src/components/custom-widget/ko/runtime/customWidget.ts
```diff
@@ -14,13 +14,19 @@ import template from "./customWidget.html";
...
+    public iframeSandboxAllows: string = iframeSandboxAllows;
     private windowRef = window;
 
     constructor() {
         this.src = ko.observable();
         this.name = ko.observable();
         this.instanceId = ko.observable();
+
+        const iframe = document.getElementsByTagName("iframe")[0];
+        this.iframeSandboxAllows = (iframeSandboxAllows + " " + iframeSandboxAllowsBrowserSpecific)
```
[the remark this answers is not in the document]
JMach1: done and what test?

ygrik (approved):

PR #2517, opened 2024-06-20
FUI - Operation details widget

jsorohova:


src/apim.runtime.module.ts
```diff
@@ -92,9 +92,10 @@ import { TraceClick } from "./bindingHandlers/traceClick";
 import { GraphqlService } from "./services/graphqlService";
 import { ListOfApisRuntimeModule } from "./components/apis/list-of-apis/listOfApis.runtime.module";
 import { ProductListRuntimeModule } from "./components/products/product-list/productList.runtime.module";
-import { ProductApisRuntimeModule } from "./components/products/product-apis/productApis.runtime.module";
 import { OperationListRuntimeModule } from "./components/operations/operation-list/operationList.runtime.module";
 import { DetailsOfApiRuntimeModule } from "./components/apis/details-of-api/detailsOfApi.runtime.module";
+import { OperationDetailsRuntimeModule } from "./components/operations/operation-details/operationDetails.runtime.module";
+import { ProductApisRuntimeModule } from "./components/products/product-apis/productApis.runtime.module";
```
JMach1, line 98: I don't see you binding this module, am I missing something?

src/components/operations/operation-details/ko/runtime/type-definition-object.html
```diff
@@ -66,9 +66,7 @@
 
                 <!-- /ko -->
             </div>
-            <!-- ko if: $component.examples -->
-            <div role="cell" class="text-truncate" data-bind="text: $component.examples[property.name], attr: {title: $component.examples[property.name]}"></div>
-            <!-- /ko -->
+            <div role="cell" class="text-truncate" data-bind="text: property.example, attr: {title: property.example}"></div>
```
JMach1, line 69: why was this removed?

src/apim.runtime.module.ts
jsorohova: It is just moved from line 95.

src/components/operations/operation-details/ko/runtime/type-definition-object.html
jsorohova: It is not removed, it is fixed. The previous logic was weird - values were taken from the big example by field name, now I am taking examples straight for the property.

src/apim.runtime.module.ts
JMach1: ooh got it

src/components/operations/operation-details/ko/runtime/type-definition-object.html
JMach1: but (from the code) if examples are not defined, we are rendering a div, while previously we were not at all, right?
I didn't check it rendered, I'm just thinking that it might look weird if there are no examples but if you rested it, it's all good

src/components/operations/operation-details/react/runtime/OperationDetails.tsx
```diff
@@ -0,0 +1,402 @@
...
+            hostnames?.length > 0 && setHostnames(hostnames);
+        });
+        loadOperation().then(loadedValues => {
+            setOperation(loadedValues.operation);
+            setTags(loadedValues.tags);
+            setDefinitions(loadedValues.definitions);
+            setRequest(loadedValues.operation?.request);
+            setResponses(loadedValues.operation?.getMeaningfulResponses());
+        }).finally(() => {
+            setWorking(false);
```
JMach1, line 83: I don't think this works as expected? what if loadApi takes longer than loadOperation?

maybe use Promise.all?

Also maybe at least some basic error handling?

src/components/operations/operation-details/react/runtime/OperationDetails.tsx
```diff
@@ -0,0 +1,402 @@
...
+
+    const loadOperation = async (): Promise<{operation: Operation, tags: Tag[], definitions: TypeDefinition[]}> => {
+        let operation: Operation;
+        let tags: Tag[];
+        let definitions: TypeDefinition[];
+
+        try {
+            if (operationName) {
+                operation = await apiService.getOperation(`apis/${apiName}/operations/${operationName}`);
+                tags = await apiService.getOperationTags(`apis/${apiName}/operations/${operationName}`);
```
JMach1, line 116: maybe make it parallel? 
and only loadDefinitions after await Promise.all

src/components/operations/operation-details/react/runtime/OperationDetailsGql.tsx
```diff
@@ -0,0 +1,172 @@
...
+        }
+        
+        setWorking(true);
+        loadApi().then(loadedApi => setApi(loadedApi));
+        loadGatewayInfo().then(hostnames => {
+            hostnames.length > 0 && setHostnames(hostnames);
+        });
+        loadGraph()
+            .then(graph => setGraph(graph))
+            .finally(() => setWorking(false));
```
JMach1, line 49: same as before with Promise.all

example of my usage:

```
const configsPromise = loadCustomWidgetConfigs(this.blobStorage, this.viewManager);
const refreshConfigs = listConfigBlobs(this.blobStorage);
Promise.all([refreshConfigs, configsPromise])
      .then(([configBlobs, configsAll]) => {
....
```

src/components/operations/operation-details/react/runtime/TypeDefinitions.tsx
```diff
@@ -0,0 +1,192 @@
...
+                                    <Body1 block className={"truncate-text"}>
+                                        <Link href={getReferenceUrl(combinationName)} title={combinationName}>{combinationName}</Link>
+                                    </Body1>
+                                );
+                            } else {
+                                children.push(<Body1 block title={combinationName}>{combinationName}</Body1>);
+                            }                                
+                        });
+                        
+                        type = <>{children}</>;
```
JMach1, line 165: why not just children?

type = children;

JMach1 (changes requested):

src/components/operations/operation-details/react/runtime/TypeDefinitions.tsx
jsorohova: Children are multiple tags and I need to wrap them into a singular one.

JMach1: I think just changing TS type of variable "type" should do the same but just a nitpick, feel free to leave it as is

PR #2536, opened 2024-07-22
Create config.yml

annaji-msft:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

Do not allow issues being created without using the issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false


ygrik (approved):

kevinpearce:
The repository readme states to use the Issues section for Feature Requests here : https://github.com/Azure/api-management-developer-portal/blob/master/README.md?plain=1#L13

The Issue form (which users will now be encouraged to use) still directs users towards the Azure Feedback Forum which is totally useless because of the current SPAM load and years old requests. 

Please update Readme and ensure forum is fixed + provide users with an alternate request route until that happens. 

Or...... just use GitHub and add a Feature Request Issue template?

annaji-msft:
@kevinpearce thank you for providing feedback. I will update the readme and also include the new feature requests should now go to discussions section. Where we could also encourage voting for the top requested feature to be picked up. https://github.com/Azure/api-management-developer-portal/discussions/categories/feature-request

PR #2661, opened 2024-08-19
Multiple fixes for Site menu item modal

jsorohova:
Included fixes:
- Correct parent displaying for all item types (it was not working for Media items)
- Removed the item itself from the list of possible parents
- Fixed error handling on item type change

### DIFF
--- a/src/admin/navigation/navigationItemModal.tsx
+++ b/src/admin/navigation/navigationItemModal.tsx
@@ -293,6 +293,8 @@ export class NavigationItemModal extends React.Component<NavigationItemModalProp
 
     processNavItemsForDropdown = (navItems: NavigationItemContract[], dropdownItems: IDropdownOption[], currentNodeLabel: string = ''): IDropdownOption[] => {
         navItems.forEach(navItem => {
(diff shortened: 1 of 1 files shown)


src/admin/navigation/navigationItemModal.tsx
```diff
@@ -624,7 +624,16 @@ export class NavigationItemModal extends React.Component<NavigationItemModalProp
                             label="Select method"
                             options={urlTypeOptions}
                             selectedKey={this.state.selectedUrlType}
-                            onChange={(event, option) => this.setState({ selectedUrlType: option.key })}
+                            onChange={(event, option) => {
+                                let errors = this.state.errors;
+                                if (option.key.toString() === LinkOptionKey.SavedUrl) {
```
ygrik, line 629: Can option (and key) parameter undefined or null?

jsorohova: No. It is a radio select with two options and one of them is always selected by default.

ygrik (approved):

PR #2669, opened 2024-08-29
Code review fix for validator

jsorohova:


rkolesnikovDX (approved):

src/admin/utils/validator.ts
rkolesnikovDX, line 1: Maybe it should be like:
``` 
(customValidation == undefined || customValidation === true)
```

rkolesnikovDX (changes requested):

rkolesnikovDX (approved):

PR #2672, opened 2024-09-11
FUI - Reports page

jsorohova:


src/components/reports/react/Charts/ApiResponseTime.tsx
```diff
@@ -0,0 +1,66 @@
...
+        {
+            legend: "Maximum response time",
+            data: []
+        }
+    ]
+}
+
+export const ApiResponseTime = ({ reportsByTime, timeRange, dateFormattingFunc }: TReportsChartProps) => {
+    const [data, setData] = useState<IChartProps>(apiResponseTimeLineChartData);
+    const [forceRerender, setForceRerender] = useState<number>(1);
```
mpodriezov, line 28: How that forceRender is used?

mpodriezov (approved):

src/components/reports/react/Charts/ApiResponseTime.tsx
jsorohova: State change checks only "top-level values". In this case it doesn't trigger re-render because lineChartData's array inner array is changed. The top-level values and structure stays the same. The state doesn't check third-level values changes and thinks that the state is not changed. This is why the forceRerender is used.

PR #2677, opened 2024-09-19
FUI - Confirm password page

jsorohova:


src/components/users/confirm-password/react/runtime/ConfirmPasswordRuntime.tsx
```diff
@@ -0,0 +1,111 @@
...
+        const values = { password, passwordConfirmation };
+        const clientErrors = validateBasic(values, validationGroup);
+
+        if (clientErrors.length > 0) {
+            dispatchErrors(this.eventManager, ErrorSources.confirmpassword, clientErrors);
+            return false;
+        }
+
+        try {
+            dispatchErrors(this.eventManager, ErrorSources.changepassword, []);
```
rkolesnikovDX, line 94: Why do we need to dispatch `changePassword` errors here?

jsorohova: Fixed. Thanks!

mpodriezov (approved):

PR #2734, opened 2024-12-09
Fix Captcha styles, fix subscribe to product alignment, prevent custom styles from interfering with predefined widgets

Onatolich:


Onatolich:
@microsoft-github-policy-service agree company="Microsoft"

src/components/products/product-subscribe/ko/runtime/product-subscribe.html
```diff
@@ -22,11 +22,11 @@
 
 <!-- ko ifnot: limitReached -->
 <form>
-    <div class="table">        
+    <div class="table">
         <div class="table-row">
             <!-- ko ifnot: delegationEnabled -->
-            <div class="col-5">
-                <input type="text" class="form-control" placeholder="Your new product subscription name" data-bind="textInput: subscriptionName" />
+            <div class="col-5 form-inline">
```
jsorohova, line 28: We don't need to change styles of the existing experience. Not sure how all existing customer would react so let's not do this.

src/themes/website/styles/widgets/widgets.scss
```diff
@@ -18,3 +18,29 @@
...
+      margin: 0 0 0.5em !important;
+      padding: 0 !important;
+    }
+
+    .form-control {
+      margin: 0 !important;
+      padding: 7px 10px 5px !important;
+    }
+  }
+}
```
jsorohova, line 46: Please, move all this to another file. Let's keep this for just imports.

Onatolich: Why? This seems as an appropriate file to import all specific widget styles as well as defining global widget styles as well.

src/themes/website/styles/widgets/widgets.scss
```diff
@@ -18,3 +18,29 @@
...
 @import "fui/reports.scss";
+
+/**
+ * If you need to make sure that custom defined styles do not interfere with widgets,
+ * this is a good place to do it. Make sure to add !important to insure overriding.
+ */
+[data-bind="widget: widget"] {
+  .button {
+    margin: 10px 20px 10px 0 !important;
+    padding: 9px 50px !important;
```
jsorohova, line 29: 50px seems like a very big number. Are you sure?

Onatolich: Just checked again, it is indeed 50px
image

src/themes/website/styles/widgets/widgets.scss
jsorohova: This file is not redesign specific. Please, create a new one and place it there.
Also, why .fui-FluentProvider is specified 3 times? Can we specify it once as old widgets don't have this class at all.

src/themes/website/styles/widgets/widgets.scss
```diff
@@ -18,3 +18,31 @@
...
+/**
+ * If you need to make sure that custom defined styles do not interfere with widgets,
+ * this is a good place to do it.
+ * Note: selector is duplicated to increase its weight. Do not use !important as it will break
+ * some widgets that rely on inline styles.
+ */
+.fui-FluentProvider.fui-FluentProvider.fui-FluentProvider {
+  .button {
+    margin: 10px 20px 10px 0;
+    padding: 9px 50px;
```
jsorohova, line 31: I am still sure this 50px value is wrong for the redesign. I know you took it from old widgets. Please, check designs and come up with proper spacing. For everything you specify here, not just this padding.

jsorohova (changes requested):

src/themes/website/styles/widgets/widgets.scss
Onatolich: > This file is not redesign specific. Please, create a new one and place it there.

Where do we put `redesign specific` files?

> Also, why .fui-FluentProvider is specified 3 times?

Read a comment I written to it

jsorohova: > Where do we put `redesign specific` files?
There is a folder named _fui_ right next to this file.

> Read a comment I written to it
Ok. My  bad. Let's leave like this.

PR #2735, opened 2024-12-11
Fixed filtering by tag for apis widget

ygrik:


src/components/apis/list-of-apis/react/runtime/ApiListRuntime.tsx
```diff
@@ -12,6 +12,7 @@ import { FiltersPosition } from "../../listOfApisContract";
 import { TApisData } from "./utils";
 import { ApiListTableCards } from "./ApiListTableCards";
 import { ApiListDropdown } from "./ApiListDropdown";
+import { Tag } from "../../../../../models/tag";
```
Onatolich, line 15: Any reason we can't add alias for src root so this import will be `@/models/tag`?

src/components/utils/react/TableFilters.tsx
```diff
@@ -49,21 +49,18 @@ export const TableFilter = ({
...
-                        value={item.value}
-                        label={item.label}
-                        checked={
-                            filtersActive[filter.value]?.includes(item.value) ??
-                            false
-                        }
+                        key={filter.id + item.id}
+                        value={item.id}
+                        label={item.name}
+                        checked={filtersActive[filter.id]?.includes(item) ?? false}
```
Onatolich, line 55: nit: `!!filtersActive[filter.id]?.includes(item)` will be shorter. But `filtersActive[filter.id]?.includes(item)` should also work fine since undefined is also an option?

Onatolich (approved):

src/components/apis/list-of-apis/react/runtime/ApiListRuntime.tsx
ygrik: Good improvement but not related to this fix.
Can be implemented later in a dedicated PR

jsorohova (approved):

PR #2755, opened 2025-01-22
Fixed admin panels not being hidden on preview mode toggle

jsorohova:
Problem:
There is a preview mode which can be toggled by pressing Ctrl+F10. The new redesigned admin panels were not hiding on this preview mode enabling.

Solution:
Added a listener for keyboard presses and admin panels visibility toggle when the mode is changing.

### DIFF
--- a/src/admin/rightPanel.tsx
+++ b/src/admin/rightPanel.tsx
@@ -99,6 +99,7 @@ export class RightPanel extends React.Component<{}, RightPanelState> {
         this.eventManager.addEventListener('onDataChange', this.onDataChange.bind(this));
         window.addEventListener('resize', this.checkScreenSize.bind(this));
         this.router.addRouteChangeListener(() => this.getPageName());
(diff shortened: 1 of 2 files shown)


src/admin/rightPanel.tsx
```diff
@@ -99,6 +99,7 @@ export class RightPanel extends React.Component<{}, RightPanelState> {
         this.eventManager.addEventListener('onDataChange', this.onDataChange.bind(this));
         window.addEventListener('resize', this.checkScreenSize.bind(this));
         this.router.addRouteChangeListener(() => this.getPageName());
+        document.addEventListener('keydown', this.onModeChange.bind(this));
```
azaslonov, line 102: Jelena, I think this is overkill, too many events. Let me still add specific event in the core. That would still make sense to have it (and a few other) events that could help hinting the editors.

jsorohova: Then I will close this PR and create a new one when the core part is ready.

PR #2869, opened 2026-03-16
fix(authService): always use redirect URI in lowercase in Auth Code with PKCE flow

### LINKED ISSUE #2868: Auth Code with PKCE Flow does not work when the Auth Server contains uppercase in its name / Id
> [!IMPORTANT]
> **PLEASE DO NOT CREATE ISSUES RELATED TO MANAGED DEVELOPER PORTAL IN THIS REPOSITORY**. Issues related to "Managed developer portal" will be closed automatically. Managed developer portal issues should be created directly in Azure Support + Troubleshooting in Azure portal. Contact Azure support (or) submit a post [on Stack Overflow](http://aka.ms/apimso). Feature requests can be raised [on the Azure Feedback Forum](https://aka.ms/apimwish).
 
Every self-hosted developer bug report should have precise description and reproduction steps; console traces (or) source code references are appreciated.

## Bug description

When the OAuth 2.0 Auth Server for the Developern Portal contains upper case letters in its name/Id the Auth Code with PKCE Flow does not work because of insistent Redirect URIs casing. (First request uses lowercase, second uppercase). Because of that the Entra ID and potential every other IDP rejects the login request because the redirect URIs do not match.

## Reproduction steps

1. Go to in Azure Portal go to APIM > Deverloper Portal > OAuth 2.0 + OpenID Connect
2. Click on 'Add'
3. Create an OAuth 2.0 Auth Server with uppercase in the Id. For example: "OAuth2".
4. Assign the Auth Server to an API
5. Go to Developer Portal and try out the API
6. Select "authorization_code (PKCE)" as the Authorization Flow
7. Pop Up appears which tries to login the user
8. Login fails because redirect URI for the first (auth) request is not the same as for the second (token) request. For example, with Entra ID the following Error is shown:

{"error":"invalid_client","error_description":"AADSTS500112: The reply address 'https://api-portal.example.com/signin-oauth/code-pkce/callback/OAuth2' does not match the reply address 'https://api-portal.example.com/signin-oauth/code-pkce/callback/oauth2' provided when requesting Authorization code.}

## Expected behavior

Correctly login the use and get an Access Token.

## Is self-hosted portal?

No (and Yes) 
> [!WARNING] 
> "Managed developer portal issues" will be closed automatically. See above for more details. 

2.34.0

N/A

## Environment

N/A

## Additional context

For first request makes URL lowercase: https://github.com/Azure/api-management-developer-portal/blob/master/src/services/oauthService.ts#L197

But second request does not (but it should too):  https://github.com/Azure/api-management-developer-portal/blob/master/src/services/oauthService.ts#L213

J0F3:
fixes #2868

### DIFF
--- a/src/services/oauthService.ts
+++ b/src/services/oauthService.ts
@@ -210,7 +210,7 @@ export class OAuthService {
                 client_id: authorizationServer.clientId,
                 code_verifier: sessionStorage.getItem("code_verifier"),
                 grant_type: GrantTypes.authorizationCode,
-                redirect_uri: redirectUri,
+                redirect_uri: redirectUri.toLowerCase(),
                 code: authorizationCode
             });
 


J0F3:
This is now open for approx 3 Month. Any updates?

